PR #354, opened 2023-07-18
perf(proxy-wasm) move a heap allocation to the stack

casimiro:
Given how small this "refactor" is I'm really unsure if it's a refactor or simply a chore.


thibaultcha:
Maybe it's actually a `perf`!

casimiro:
oh that's right! `perf` it is. will update the commit message and re-push the commit.

PR #361, opened 2023-07-23
docs(user) remove zig proxy-wasm SDK link

flrgh:
[PROXY_WASM.md](/Kong/ngx_wasm_module/blob/main/docs/PROXY_WASM.md) currently links out to [mathetake/proxy-wasm-zig-sdk](https://github.com/mathetake/proxy-wasm-zig-sdk/) for the Zig SDK, but this repo appears to have been removed by its author, who has also [removed its link from the proxy-wasm spec README](https://github.com/proxy-wasm/spec/commit/d2c545ba2d09a1ffa303dedb6626021347d23f17).

Of note, I found what looks to be somebody's [unofficial mirror](https://github.com/iCodeIN/proxy-wasm-zig-sdk) of this repo, but it's likely to be considered unmaintained, so I think we should just remove Zig altogether.


thibaultcha:
Sounds good. There is another Zig link in the root README.md, can you remove it as well?

thibaultcha:
Merged!

PR #362, opened 2023-07-25
fix(wrt/wasmtime) call wasmtime functions after fork

casimiro:
When configured to run as a daemon, Nginx will fork itself right after parsing its configuration.

Before this commit `ngx_wavm_init` was being called during config initialization, i.e. before Nginx fork.

Because of this, Wasmtime C API `wasm_engine_new_with_config` function was being called before Nginx would fork itself while the remaining Wasmtime API calls would be invoked in the newly forked process.

Unfortunately, this use case is not well handled by Wasmtime, leading it to panic when `wasmtime_func_call` is called.

This commit moves `ngx_wavm_init` invocation from `ngx_wasm_core_init` to `ngx_wasm_core_init_process`. Which causes `ngx_wavm_engine_init` to be called after Nginx would fork itself.

### DIFF
--- a/t/01-wasm/directives/001-module_directive.t
+++ b/t/01-wasm/directives/001-module_directive.t
@@ -174,7 +174,7 @@ qr/\[emerg\] .*? \[wasm\] open\(\) ".*?none\.wat" failed \(2: No such file or di
 ]
 --- no_error_log
 [error]
---- must_die
+--- must_die: 0
 
 
 
@@ -236,7 +236,7 @@ qr/\[emerg\] .*? \[wasm\] open\(\) ".*?none\.wat" failed \(2: No such file or di
 ]
 --- no_error_log
 [error]
---- must_die
+--- must_die: 0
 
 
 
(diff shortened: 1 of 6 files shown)


t/01-wasm/directives/001-module_directive.t
```diff
@@ -136,7 +136,7 @@ qr/\[emerg\] .*? \[wasm\] open\(\) ".*?none\.wat" failed \(2: No such file or di
 --- no_error_log
 [error]
 [crit]
---- must_die
+--- must_die: 0
```
thibaultcha, line 139: Btw, we cannot go ahead with these changes. We *need* the master process to have a Wasm VM and detect invalid modules/bytecode, or else we start worker processes with an invalid configuration. This behavior is on purpose: it is so that we catch configuration errors and prevent Nginx from starting, just like configuration parsing also prevents Nginx from starting.

thibaultcha: I think the solution we should dig, which should also address other concerns and problems we have with the module, is that the master process should start a temporary VM that does all this, and then stop it before the `fork()`. Then forked worker processes restart the VM, assuming all configuration is valid (because checked by the master process).

This is actually the original intended design of this module ^. I've implemented all this maybe a couple of years ago, and there were changes with the V8 integration as well, I'm not sure how we stand with regards to this intended behavior today, and cannot check right moment.

PR #365, opened 2023-07-28
chore(tests) enable rust backtrace

casimiro:
Add `env RUST_BACKTRACE=full;` to nginx.conf

### DIFF
--- a/t/03-proxy_wasm/003-on_tick.t
+++ b/t/03-proxy_wasm/003-on_tick.t
@@ -128,6 +128,7 @@ ok
 
 
 === TEST 5: proxy_wasm - multiple ticks
+--- disable_rust_backtrace
 'daemon off' must be set to check exit_code is 2
 Valgrind mode already writes 'daemon off'
 HUP mode does not catch the worker exit_code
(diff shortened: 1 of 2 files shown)


t/03-proxy_wasm/003-on_tick.t
```diff
@@ -128,6 +128,7 @@ ok
 
 
 === TEST 5: proxy_wasm - multiple ticks
+--- disable_rust_backtrace
```
thibaultcha, line 131: Can't we update the error log regex and always have the backtrace details env directives instead?

casimiro: I'm not sure. It's not the error log regex that's conflicting with the backtrace, it seems to be the `stub`:
`pattern "stub" should not match any line in error.log but matches line "   7: wasmtime::trampoline::func::stub_fn" (req 0)`

It also seemed right suppressing the backtrace from the console output in a successful test run.

thibaultcha:
For now I think we should only focus on the Wasmtime crash anyway.

PR #371, opened 2023-08-02
what

thibaultcha:
Trap format changed without the runtimes versions changing? 🤔  Sounds like a Rust dep change.


thibaultcha:
Yes, confirmed it's a Rust update change. Closing for another PR to address this.

PR #374, opened 2023-08-04
feat(*) wasm_socket unix domain socket support

casimiro:
- [x] Update wasm_socket port parsing
- [x] Add test cases (sanity, missing socket file, missing :authority)
- [x] ~Update tests relying `TEST_NGINX_SERVER_PORT2` to use unix domain socket instead~. 
- [x] ~If possible, drop `TEST_NGINX_SERVER_PORT2`~

Updating tests that were using `TEST_NGINX_SERVER_PORT2` ended up being a substantial amount of work on it's own, deserving it's own PR or PRs.

### DIFF
--- a/src/common/ngx_wasm_socket_tcp.c
+++ b/src/common/ngx_wasm_socket_tcp.c
@@ -119,9 +119,11 @@ ngx_wasm_socket_tcp_resume(ngx_wasm_socket_tcp_t *sock)
 
 ngx_int_t
 ngx_wasm_socket_tcp_init(ngx_wasm_socket_tcp_t *sock,
-    ngx_str_t *host, in_port_t port, unsigned tls,
-    ngx_wasm_subsys_env_t *env)
+    ngx_str_t *host, unsigned tls, ngx_wasm_subsys_env_t *env)
 {
+    u_char            *p, *last;
+    static ngx_str_t   uds_prefix = ngx_string("unix:");
+
     ngx_memzero(sock, sizeof(ngx_wasm_socket_tcp_t));
 
     ngx_memcpy(&sock->env, env, sizeof(ngx_wasm_subsys_env_t));
@@ -165,7 +167,40 @@ ngx_wasm_socket_tcp_init(ngx_wasm_socket_tcp_t *sock,
     sock->url.default_port = 80;
 #endif
     sock->url.url = sock->host;
-    sock->url.port = port;
+    sock->url.port = 0;
+
+    if (host->len > uds_prefix.len
+        && ngx_memcmp(host->data, uds_prefix.data, uds_prefix.len) == 0)
+    {
+
+#if (!NGX_HAVE_UNIX_DOMAIN)
+        ngx_wasm_log_error(NGX_LOG_ERR, sock->log, 0,
+                           "host \"%V\" requires unix domain socket support",
+                           host);
+        return NGX_ERROR;
+#endif
+
+    } else {
+        /* extract port */
+
+        p = host->data;
+        last = p + host->len;
+
+        if (*p == '[') {
+            /* IPv6 */
+            p = ngx_strlchr(p, last, ']');
+            if (p == NULL) {
+                p = host->data;
+            }
+        }
+
+        p = ngx_strlchr(p, last, ':');
+
+        if (p) {
+            sock->url.port = ngx_atoi(p + 1, last - p);
+        }
+    }
+
     sock->url.no_resolve = 1;
 
     if (ngx_parse_url(sock->pool, &sock->url) != NGX_OK) {
@@ -207,6 +242,8 @@ ngx_wasm_socket_tcp_connect(ngx_wasm_socket_tcp_t *sock)
     ngx_stream_core_srv_conf_t           *ssrvcf;
     ngx_stream_session_t                 *s;
 #endif
+
+    dd("enter");
 
     if (sock->errlen) {
         return NGX_ERROR;
@@ -284,7 +321,15 @@ ngx_wasm_socket_tcp_connect(ngx_wasm_socket_tcp_t *sock)
                        "wasm tcp socket no resolving: %V",
                        &sock->resolved.host);
 
-        return ngx_wasm_socket_tcp_connect_peer(sock);
+        rc = ngx_wasm_socket_tcp_connect_peer(sock);
+        if (sock->connected) {
+#if (NGX_SSL)
+            if (sock->ssl_conf) {
+                return ngx_wasm_socket_tcp_ssl_handshake(sock);
+            }
+#endif
+        }
+        return rc;
     }
 
     ngx_log_debug1(NGX_LOG_DEBUG_WASM, sock->log, 0,
@@ -1484,7 +1529,7 @@ ngx_wasm_socket_tcp_init_addr_text(ngx_peer_connection_t *pc)
         break;
 #endif
 
-#if (NGX_HAVE_UNIX_DOMAIN && 0)
+#if (NGX_HAVE_UNIX_DOMAIN)
     case AF_UNIX:
         addr_text_max_len = NGX_UNIX_ADDRSTRLEN;
         break;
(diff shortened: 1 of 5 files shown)


src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -150,6 +150,9 @@ ngx_http_proxy_wasm_dispatch(ngx_proxy_wasm_exec_t *pwexec,
     ngx_http_wasm_req_ctx_t         *rctxp = NULL;
     ngx_http_proxy_wasm_dispatch_t  *call = NULL;
     unsigned                         enable_ssl = 0;
+#if (NGX_HAVE_UNIX_DOMAIN)
+    static ngx_str_t                 uds_prefix = ngx_string("unix:");
+#endif
```
casimiro, line 156: @thibaultcha I'm uncertain where UDS related variables should be declared.

`uds_prefix` is used both by `ngx_http_proxy_wasm_dispatch` and `ngx_http_proxy_wasm_dispatch_request` which makes me want to declare it at the beginning of the file, maybe after `static char  ngx_http_proxy_wasm_host[] = "Host: ";`.

I would like declare it as a `ngx_str_t` to benefit from the `len` attribute later on, but since the other globals that would precede it are actually `char` arrays, it doesn't sound like a good fit either.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -606,7 +621,24 @@ ngx_http_proxy_wasm_dispatch_request(ngx_http_proxy_wasm_dispatch_t *call)
 
     b->last = ngx_cpymem(b->last, ngx_http_proxy_wasm_host,
                          sizeof(ngx_http_proxy_wasm_host) - 1);
+
+#if (NGX_HAVE_UNIX_DOMAIN)
+    static ngx_str_t uds = ngx_string("unix:");
+    static ngx_str_t localhost = ngx_string("localhost");
```
thibaultcha, line 627: There is nothing that guarantees `localhost` is a valid hostname, it may not be in `/etc/hosts`. I think we should copy the authority as it was given by the user; the only difference with `unix` socket is that `port` must be set to `0`.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -606,7 +621,24 @@ ngx_http_proxy_wasm_dispatch_request(ngx_http_proxy_wasm_dispatch_t *call)
...
+
+    if (call->authority.len > uds.len
+        && ngx_str_eq(call->authority.data, uds.len, uds.data, uds.len))
```
thibaultcha, line 630: This is not proper usage of `ngx_str_eq`, this would be:
```c
ngx_str_eq(call->authority.data, call->authority.len, uds.data, uds.len)
```

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -606,7 +621,24 @@ ngx_http_proxy_wasm_dispatch_request(ngx_http_proxy_wasm_dispatch_t *call)
...
+    {
+        b->last = ngx_cpymem(b->last, localhost.data, localhost.len);
+
+    } else {
+        b->last = ngx_cpymem(b->last, call->authority.data,
```
thibaultcha, line 635: There are better ways to use preprocessor directives that would avoid duplicating this line.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -240,11 +243,23 @@ ngx_http_proxy_wasm_dispatch(ngx_proxy_wasm_exec_t *pwexec,
...
     }
 
+#if (NGX_HAVE_UNIX_DOMAIN)
+    if (host->len < uds_prefix.len ||
+        !ngx_str_eq(p, uds_prefix.len, uds_prefix.data, uds_prefix.len))
+    {
+        p = ngx_strlchr(p, last, ':');
+
+        if (p) {
+            port = ngx_atoi(p + 1, last - p);
```
thibaultcha, line 253: There are better ways to use preprocessor directives that would avoid duplicating this port-extraction logic, all we need to do is not run this logic if we have `NGX_HAVE_UNIX_DOMAIN` and the `unix:` protocol in the host.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -240,11 +243,23 @@ ngx_http_proxy_wasm_dispatch(ngx_proxy_wasm_exec_t *pwexec,
         }
     }
 
+#if (NGX_HAVE_UNIX_DOMAIN)
+    if (host->len < uds_prefix.len ||
+        !ngx_str_eq(p, uds_prefix.len, uds_prefix.data, uds_prefix.len))
```
thibaultcha, line 248: Same here, this feels like it could be a single `ngx_str_eq`, without the length comparison above. Reminder that `ngx_str_eq` already checks for lengths matching before doing any comparison of bytes.

thibaultcha (commented):
Overall I am weary that our `host` parsing logic is not integrated within `ngx_wasm_sock`, which is what I had in mind for this feature... I know the dispatch caller already does some parsing ahead of delegating to `ngx_wasm_sock`, but moving more parsing logic to `ngx_wasm_sock` (or validating logic) is also something I would have liked to be explored.

t/03-proxy_wasm/hfuncs/130-proxy_dispatch_http.t
```diff
@@ -263,7 +263,46 @@ Hello back
 
 
 
-=== TEST 14: proxy_wasm - dispatch_http_call() resolver + hostname (IPv4)
+=== TEST 14: proxy_wasm - dispatch_http_call() unix domain socket
```
thibaultcha, line 266: We need more than 1 test; e.g. invalid host and/or non-existing socket

casimiro:
Thank you for the detailed review @thibaultcha! I'll change the code to address all comments.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -227,25 +225,6 @@ ngx_http_proxy_wasm_dispatch(ngx_proxy_wasm_exec_t *pwexec,
...
-        if (p == NULL) {
-            p = host->data;
-        }
-    }
-
-    p = ngx_strlchr(p, last, ':');
-
-    if (p) {
-        port = ngx_atoi(p + 1, last - p);
-    }
```
thibaultcha, line 247: I think it's totally a win that this moved to ngx_wasm_sock 👍🏼

src/common/ngx_wasm_socket_tcp.c
```diff
@@ -284,7 +319,15 @@ ngx_wasm_socket_tcp_connect(ngx_wasm_socket_tcp_t *sock)
                        "wasm tcp socket no resolving: %V",
                        &sock->resolved.host);
 
-        return ngx_wasm_socket_tcp_connect_peer(sock);
+        rc = ngx_wasm_socket_tcp_connect_peer(sock);
+        if (sock->connected) {
+#if (NGX_SSL)
+            if (sock->ssl_conf) {
+                return ngx_wasm_socket_tcp_ssl_handshake(sock);
```
thibaultcha, line 326: Did you find some kind of problem with SSL handshakes? This sort of catches my eye... And now I'm wondering if we don't have some problems with triggering SSL handshakes?

casimiro: Yes I did.

I noticed the SSL handshake was not being performed when the dispatched request targeted a unix domain socket.

And that was happening because when the request was dispatched, `ngx_wasm_socket_tcp_connect` ran only once due to `ngx_wasm_socket_tcp_connect_peer` returning `NGX_OK` rather than `NGX_AGAIN`.

This prevented `ngx_wasm_socket_tcp_ssl_handshake` from ever being invoked [here](https://github.com/Kong/ngx_wasm_module/blob/main/src/common/ngx_wasm_socket_tcp.c#L215-L222) since the first time `ngx_wasm_socket_tcp_connect` is called `sock->connected` would still be false.

I was indeed going to ping you over this change as I'm not sure this is the solution we want for this behavior.

thibaultcha: Hmm so this means when specifying an IP address (not `unix`, and no DNS resolution) we are also not initiating an SSL handshake... We will need tests for this use-case as well!

PR #381, opened 2023-08-10
 feat(proxy-wasm) implement response body buffering

thibaultcha:
First this fixes `on_response_body` execution of requests when they issue subrequests that produce a body (see 004-on_http_phases.t). This unlocks testing of response body buffering via subrequests producing chunked responses (`ngx_chain_t` buffers).

Response body buffering itself is implemented as part of ngx_http_wasm_filter_module, so as to be usable from other components than just proxy-wasm. Response body buffering is enabled via `rctx->resp_buffering` which is enabled when `ngx_wasm_ops_resume` returns `NGX_AGAIN`, which for filters translates to returning `PAUSE` from `on_http_response_body`.

If and when the response body buffers are full, then the next `on_http_response_body` call will have `eof=false`, and more invocations can be expected, as the body *must* be proxied in chunks.

If the response body fits in the buffers, then the next `on_http_response_body` call will have `eof=true` as one would expect.

We catch "response body buffering already requested" in ngx_proxy_wasm to produce a proxy-wasm error message, as this feature is currently only exposed through it.

---

- [x] buffer reuse
- [ ] ~gzip tests~ - out of scope (dev env binary work)
- [x] CI & Valgrind
- [x] coverage
- [x] documentation
- [ ] ~changelog~ - out of scope


src/http/ngx_http_wasm_filter_module.c
```diff
@@ -179,3 +225,163 @@ ngx_http_wasm_body_filter_handler(ngx_http_request_t *r, ngx_chain_t *in)
...
+            break;
+
+        } else if (cl->buf->last_in_chain) {
+            if (rctx->r != rctx->r->main) {
+                rctx->resp_chunk_eof = 1;
+            }
+
+            break;
+        }
+    }
```
thibaultcha, line 253: This loop is the last thing that I would like to improve before merging this, because it may have impact into how we may reuse chained links in the future; this loop attaches buffers that do not belong to ngx_wasm_module, to the response body buffers, which do belong to ngx_wasm_module. Sometimes, modules don't set the buffer's `tag` (e.g. echo-nginx-module), so distinguishing where buffers come from becomes difficult. I'll have to mull over it and see how this can ensure buffer reuse in the future (or not, perhaps).

PR #386, opened 2023-08-17
chore(config) better support NGX_WASM_CARGO=0

hishamhm:
The existing NGX_WASM_CARGO=0 variable already allows skipping the Cargo step for building ngx-wasm-rs during `./configure`, but it didn't set the defines needed for each VM configuration.

This commit adds the defines when NGX_WASM_CARGO=0 is used, allowing the user to build ngx-wasm-rs separately if wanted, passing its header and library locations via NGX_BUILD_CC_OPTS and NGX_BUILD_LD_OPTS, or via `--cc-opts` and `--ld-opts` flags in the nginx `configure` script.

Includes tests and documentation.

(Nevermind the branch name, the approach taken changed halfway through :laughing: )

### DIFF
--- a/config
+++ b/config
@@ -92,12 +92,19 @@ have=NGX_WASM_RUNTIME value="\"$ngx_wasm_runtime_name\"" . auto/define
 
 # auto/cargo (ngx_wasm_rs)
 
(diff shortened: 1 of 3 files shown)


thibaultcha:
Merged but I moved the documentation to INSTALL.md since it is end-user facing!

PR #392, opened 2023-08-22
feat(build) add bazel rules to build ngx-wasm-rs and ngx-rust

fffonion:
This PR adds Bazel build rules to compile ngx-wasm-rs and ngx-rust that can be imported and used by Kong,
as directly calling `cargo` might break cross-compilation.

The `rules_rust` doesn't seem to parse the local redirection of depenendencies in Cargo.toml well so that
I have to manually add those dependencies. And it seems the secondary dependencies will also needs to
be in the top level Cargo.toml to make build happy. Let me know how do you feel about this workaround.

What's missing at present:
- [ ] wat feature (currently not working due to "missing CMAKE_ROOT")
- [ ] readme
- [ ] reusable workflow


thibaultcha:
@fffonion Are you prepared to be responsible for changes in the Bazel structure as development of the module itself continues? I don't feel very good about merging a whole new build system that we aren't comfortable maintaining, so someone has to be prepared to help us maintain it. Are you?

build/rust-nightly.bzl
```diff
@@ -0,0 +1,62 @@
...
+    binary_name = name + "_nightly"
+    _transition_rule(
+        name = name,
+        actual_binary = ":%s" % binary_name,
+    )
+    rust_static_library(
+        name = binary_name,
+        visibility = visibility,
+        **kwargs
+    )
```
thibaultcha, line 62: How much of this can we remove once #393 is merged?

Cargo.toml
```diff
@@ -1,4 +1,4 @@
-[workspace]
```
thibaultcha, line 1: Why removing this? Is it optional or something?

build/README.md
```diff
@@ -0,0 +1,44 @@
...
+configure_make(
+    name = "openresty",
+    # ...
+    deps = [
+        "@ngx_wasm_module",
+    ],
+)
+```
+
+When building this library in Bazel, use the `-c opt` flag to ensure optimal performance. The default fastbuild mode produces a less performant binary.
```
thibaultcha, line 44: Should `NGX_WASM_CARGO=0` be mentioned anywhere in this document?

build/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```python
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+
+git_repository(
+    name = "ngx_wasm_module",
+    branch = "some-tag",
+    remote = "https://github.com/Kong/atc-router",
```
thibaultcha, line 14: ditto

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,35 @@
...
+    - '**/*.md'
+    branches:
+    - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  tests:
```
thibaultcha, line 20:
```suggestion
  bazel-tests:
```

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,35 @@
...
+    name: Tests
```
thibaultcha, line 21:
```suggestion
    name: Bazel Tests
```

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    steps:
+    - name: Checkout source code
+      uses: actions/checkout@v3
+
+    - name: Build
+      run: bazel build :ngx_rust :ngx_wasm_rs_static :ngx_wasm_rs_shared --verbose_failures
+    
+    - name: Display artifacts
+      run: ls -lh bazel-bin/
```
thibaultcha, line 35: On second thought, we will need this to be a reusable workflow that we can invoke from `ci.yml` and `ci-large.yml`. Please follow the example with `job-unit-tests.yml` (or any of the `job-*.yml` reusable workflows we already have).

Cargo.toml
thibaultcha: I don't think this change is correct; it also breaks CI. Please revert.

build/rust-nightly.bzl
fffonion: This file can now be removed.

build/README.md
```diff
@@ -0,0 +1,44 @@
+# Bazel project for atc-router
```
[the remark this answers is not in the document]
fffonion: Updated the readme, it may still need some change on the example part, once this PR is finished.

Cargo.toml
fffonion: Yes, that was an incorrect change due to a previous hack that I missed in cleanup, reverted.

build/README.md
fffonion: fixed

build/README.md
fffonion: Good point, I have mentioned in example.

.github/workflows/bazel.yml
fffonion: I will leave this as a TODO for this PR once I finish the bazel part of it.

fffonion:
@thibaultcha This PR only intended to build the rust part of this project, but not all of the Nginx build system. And it won't replace the default cargo based build when using outside of Kong (CI, vanilla nginx). So it would need minimal maintainance once merged, the most part of work will be at Kong side. I'm happy to setup knowledge sessions with wasm team folks on this : )

config
```diff
@@ -102,7 +102,7 @@ if [ "$ngx_wasm_cargo_lib_name" = ngx_wasm_rs ]; then
         fi
     fi
 
-    if echo -n "$ngx_wasm_runtime_libs\n$NGX_LD_OPT" | grep -q "ngx-wasm-rs"; then
```
thibaultcha, line 105: This greps for the path `ngx-wasm-rs` as it is specified as a path, not a library name. Please revert.

fffonion: The path name doesn't necessarily contain `ngx-wasm-rs` (for example the rust rule automatically export the file in `lib` under build dependency https://github.com/Kong/kong-ee/pull/6346/files#diff-5c8ca1cd9c76393e97326f7857dd2be6cd741552bab8e4413dbcad5b652a94ccR29). Well, it's probably not reliable to either test the path or the module name, as it may fasly match some other part in NGX_LD_OPT; how do you feel?

thibaultcha: > The path name doesn't necessarily contain `ngx-wasm-rs`

It does when using the local development environment, hence why keeping it that way. For Bazel there will have to be another way of detecting this.

thibaultcha:
Latest remnants of CI flakiness were fixed in `main`, so a rebase should get rid of these pesty failures here too! 👍🏼

thibaultcha:
Hmmm the PR comes from your fork, it seems we still have ti adjust for that use-case. I hope #400 should fix it.

fffonion:
Thank you @thibaultcha ! I will make another rebase. I may continue to work on this PR on the TODO items shortly.

thibaultcha:
Hmmm that last failure is what I was afraid of, that the ghcr.io tag would also fetch upstream instead of just finding it locally... So that means we have to use a different tag when the `GITHUB_TOKEN` is missing...

fffonion:
@thibaultcha Is there anything I can setup on my fork to fix it? I understand secrets of GHA for fork is such a headache.

thibaultcha:
@fffonion Give it another rebase since the last update I put in there yesterday, it should work as I tested it, short of something specific I didn't foresee.

lib/ngx-wasm-rs/Cargo.toml
```diff
@@ -8,6 +8,10 @@ edition = "2018"
 ngx-wasm-wat = { path = "lib/wat", optional = true }
 ngx-wasm-backtrace = { path = "lib/backtrace" }
 
+# from backtrace
+wasmparser = "0.102"
+rustc-demangle = "0.1"
```
hishamhm, line 13: Why are these dependencies duplicated here? Cargo resolves hierarchical dependencies nicely, no need to specify this twice.

fffonion: These are the workaround for bazel's rust rule, it doesn't use `cargo` cli directly and doesn't resolve hierachical dependecies like `cargo` did 😢 .

config
```diff
@@ -102,7 +102,7 @@ if [ "$ngx_wasm_cargo_lib_name" = ngx_wasm_rs ]; then
...
+    if echo -n "$ngx_wasm_runtime_libs\n$NGX_LD_OPT" | grep -q "ngx_wasm_rs"; then
```
curiositycasualty, line 105:
```suggestion
    if echo -n "$ngx_wasm_runtime_libs\n$NGX_LD_OPT" | grep -E -q "ngx_wasm_rs|ngx-wasm-rs"; then
```

PR #407, opened 2023-09-05
fix(proxy-wasm) improve instance recycling robustness

thibaultcha:
* Allow for recycling the root instance (not managed by isolation modes)
* Allow for smoother recycling of http instances (including when the root instance has trapped)
* Also recycle trapped instances in the busy queue
* Less verbose context creation of fresh and/or recycled instances starting in atypical steps (e.g. on_response_headers).


thibaultcha:
This uncovered an issue we have with SIGHUP and pending timers, which will need to be aborted through a triggered event. WIP.

thibaultcha:
This turned into what is currently a pretty heavy refactor of the proxy-wasm instance/context relationship, which was still unfinished (had lots of leftover code from before `ictx`, which makes the handling of root/filter contexts inconsistent).

thibaultcha:
Moved to #422.

PR #409, opened 2023-09-07
chore(ci) bump and sync CC version among Dockerfiles and CI workflows

casimiro:
Dockerfiles and CI workflow use the latest version available in the runner OS.

Older versions of gcc and clang are tested in CI Large.

### DIFF
--- a/.github/workflows/job-build-tests.yml
+++ b/.github/workflows/job-build-tests.yml
@@ -1,0 +1,67 @@
+name: Build tests
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+      cc:
+        required: true
+        type: string
+      ngx:
+        required: true
+        type: string
+      runtime:
+        required: true
+        type: string
+      wasmtime:
+        required: true
+        type: string
+      wasmer:
+        required: true
+        type: string
+      v8:
+        required: true
+        type: string
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  CC: ${{ inputs.cc }}
+  NGX_WASM_RUNTIME: ${{ inputs.runtime }}
+
+jobs:
+  build:
+    name: 'Build'
+    #if: ${{ false }}
+    runs-on: ${{ inputs.os }}
+    steps:
+      - name: 'Setup Ubuntu deps - gcc'
+        if: ${{ !env.ACT && contains(inputs.os, 'ubuntu') && contains(inputs.cc, 'gcc') }}
+        run: sudo apt-get update && sudo apt-get install -y ${CC} libstdc++-${CC#*-}-dev
+
+      - name: 'Setup Ubuntu deps - clang'
+        if: ${{ !env.ACT && contains(inputs.os, 'ubuntu') && contains(inputs.cc, 'clang') }}
+        run: sudo apt-get update && sudo apt-get install -y ${CC}
+
+      - name: 'Setup macOS deps'
+        if: ${{ !env.ACT && contains(inputs.os, 'macos') }}
+        run: brew install ninja openssh
+
+      - uses: actions/checkout@v3
+      - name: 'Setup cache - work/ dir'
+        uses: actions/cache@v3
+        if: ${{ !env.ACT }}
+        with:
+          path: |
+            work/downloads
+            work/runtimes
+            work/openssl
+          key: work-${{ runner.os }}-${{ inputs.cc }}-${{ inputs.ngx }}-${{ inputs.runtime }}-${{ hashFiles('util/**/*.sh', 'util/**/*.pl', 'util/**/*.awk', '.github/**/*.yml', '.github/**/*.sh', '.github/**/*.js', 'rust-toolchain', 'Makefile') }}
+      - name: Setup Wasm runtime
+        run: ./util/runtime.sh -R ${{ inputs.runtime }} -V ${{ inputs[inputs.runtime] }}
+      - run: make setup
+      - run: make test-build
--- a/.github/workflows/job-unit-tests.yml
+++ b/.github/workflows/job-unit-tests.yml
@@ -70,8 +70,8 @@ jobs:
       coveralls_name: ${{ steps.lcov.outputs.name }}
     steps:
       - name: 'Setup deps - apt-get'
-        if: ${{ inputs.cc == 'gcc-9' }}
-        run: sudo apt-get update && sudo apt-get install -y gcc-9 libstdc++-9-dev lcov
+        if: ${{ contains(inputs.os, 'ubuntu') }}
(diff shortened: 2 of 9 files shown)


src/wasm/wrt/ngx_wrt_wasmer.c
```diff
@@ -558,7 +560,9 @@ ngx_wasmer_init_instance(ngx_wrt_instance_t *instance, ngx_wrt_store_t *store,
                     instance->env.data[i] = (wasm_extern_t *)
                         wasmer_named_extern_unwrap(
                             instance->wasi_imports.data[j]);
+#if (NGX_DEBUG)
```
casimiro, line 563: @thibaultcha is there supposed to be an empty line following these `#endif`s?

src/wasm/ngx_wasm_ops.c
```diff
@@ -126,7 +126,7 @@ ngx_wasm_ops_plan_load(ngx_wasm_ops_plan_t *plan, ngx_log_t *log)
     ngx_uint_t               *fid;
     ngx_array_t              *ids;
     ngx_wasm_op_t            *op;
-    ngx_wasm_ops_pipeline_t  *pipeline;
+    ngx_wasm_ops_pipeline_t  *pipeline = NULL;
```
thibaultcha, line 129: May I suggest making a separate commit for the `src`/ changes? That way they will appear in the changelog as fix for newer compiler versions!

src/wasm/wrt/ngx_wrt_wasmer.c
thibaultcha: If the surrounding code is already "airy", then usually I keep it that style. If it is dense then I keeping them dense...

src/wasm/ngx_wasm_ops.c
casimiro: Thank you for pointing that out! I'll make a separate commit for the src/ changes.

.github/workflows/job-unit-tests.yml
```diff
@@ -70,8 +70,8 @@ jobs:
       coveralls_name: ${{ steps.lcov.outputs.name }}
     steps:
       - name: 'Setup deps - apt-get'
-        if: ${{ inputs.cc == 'gcc-9' }}
-        run: sudo apt-get update && sudo apt-get install -y gcc-9 libstdc++-9-dev lcov
+        if: ${{ !env.ACT }}
```
thibaultcha, line 73: This is perhaps the only thing I would consider changing: with this `if` condition, if we were to bump the version number in the CI workflows, but forget to bump it in the Act image Dockerfile, the Act run would be failing since this step is skipped. If we always enable this step, at worst the `apt-get install` would be skipped (packages already installed in the Act image), and at best it would install the newer versions, and ensure the Act run continues. Basically, the Act image installation would only be an optimization (early install). Makes sense?

assets/release/Dockerfiles/Dockerfile.ubuntu-22.04
```diff
@@ -26,6 +28,7 @@ ENV PATH $CARGO_HOME/bin:$PATH
 RUN mkdir -p "$CARGO_HOME" && mkdir -p "$RUSTUP_HOME" && \
     curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable && \
     chmod -R a=rwX $CARGO_HOME $RUSTUP_HOME
+RUN rustup target add wasm32-wasi && rustup target add wasm32-unknown-unknown
```
thibaultcha, line 31: And another minor thing: can we move this to the below `RUN` step? This `RUN` step is the same for all release images; only this image has an additional `RUN` step to install CI/CD dependencies below after the "Note" comment. We can also avoid a new  `RUN` step there and do `&& \ [newline] rustup target add ...` anywhere you see fit in that chain of commands down there.

Also since the Rustup target install step in the CI is skipped on Act, it makes me wonder how come Act worked then? 🤔

.github/workflows/job-unit-tests.yml
thibaultcha: Oh, and also we should have an `os` check for this step, since once we re-enable macOS jobs it will have to be skipped.

.github/workflows/job-build-tests.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  build:
+    name: 'Build'
+    #if: ${{ false }}
+    runs-on: ${{ inputs.os }}
+    steps:
+      - name: 'Setup deps - apt-get gcc'
+        if: ${{ !env.ACT && contains(inputs.cc, 'gcc') }}
+        run: sudo apt-get update && sudo apt-get install -y ${CC} libstdc++-${CC#*-}-dev lcov
```
thibaultcha, line 44:
```suggestion
        run: sudo apt-get update && sudo apt-get install -y ${CC} libstdc++-${CC#*-}-dev
```

.github/workflows/job-build-tests.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: 'Setup deps - apt-get clang'
+        if: ${{ !env.ACT && contains(inputs.cc, 'clang') }}
+        run: sudo apt-get update && sudo apt-get install -y ${CC} lcov
```
thibaultcha, line 48:
```suggestion
        run: sudo apt-get update && sudo apt-get install -y ${CC}
```

assets/release/Dockerfiles/Dockerfile.ubuntu-22.04
casimiro: Thank you for noticing that! I will accommodate `rustup` in that last chain of commands. 

> Also since the Rustup target install step in the CI is skipped on Act, it makes me wonder how come Act worked then?

That also stroke me as odd. In a recent [commit](https://github.com/Kong/ngx_wasm_module/commit/70552dd1ac7af07fb13e70213ab2d2dc84451b25#diff-eb2b51a21328577173b0f626df0ca814350888705a59f9575e08538acf532fcf) we introduced `if: ${{ !env.ACT }}` to the steps installing rust and go in both `unit` and `valgrind` jobs. Although the change made sense as both dependencies are installed in the docker image used by act, it lacked rust targets setup.

Act was consistently failing jobs due to missing rust targets:
```
| error[E0463]: can't find crate for `core`
|   |
|   = note: the `wasm32-wasi` target may not be installed
|   = help: consider downloading the target with `rustup target add wasm32-wasi`
```

Then after including the rustup commands in the Dockerfile, jobs successfully compiled rust plugins.

.github/workflows/job-unit-tests.yml
casimiro: Indeed. Thank you for the good points.

PR #419, opened 2023-09-20
tests(*) tests for client/upstream connection drop

casimiro:
This PR proposes a set of test cases that documents the behaviour of ngx_wasm_module on the events of connection drop both from client and upstream. 

It also include fixes for bugs encountered during the investigation.


thibaultcha:
@casimiro Could you rebase this PR so we can go ahead with a smooth merge? Thanks!

thibaultcha:
@casimiro Just rebase and the Valgrind failures will go away.

PR #420, opened 2023-09-22
fix(proxy-wasm) fix memory growth on low-isolation modes

hishamhm:
We were observing linear memory growth when running requests in low-isolation modes.

The root cause comes from the fact that small allocations had instance-long lifetimes (including, for example, the arrays for `args` and `rets` for each Wasm call, causing the observable linear growth when running `wrk`).

Turns out that `ngx_pfree`, which was being used in the code to free `args` and `rets`, [does _not_ free small allocations](https://github.com/nginx/nginx/blob/027b68168867514665751a65780f050952bc48ec/src/core/ngx_palloc.c#L277-L294). It only frees large allocations; small allocations are only freed when the pool itself is destroyed.

I observed an initial improvement by moving args and rets allocations to the stack. I added this as a `perf()` commit here since I think this is still a worthwhile change, given these are usually very small arrays.

The general solution, though, is to use a request-lived pool for `pwctx`, so we don't hoard small request-specific allocations over the lifetime of an instance.

A third commit changes the setup of the `pwexec` pool so that it uses the `pwctx` pool under all isolation modes, like it does for logs. Wasm isolation mode should pertain only to the Wasm-accessible store, not to the temporary C-side data structures, which we now free on each request.

Given our new understanding of `ngx_pfree`, I think there are probably more opportunities in the codebase for replacing known-small data which is deterministically freed either with a lighter-weight `ngx_alloc`/`ngx_free` pair, or even with stack allocations, but I wanted to keep this PR straight-to-the-point and not go hunting for more changes.

Before and after memory graphs:

image
image

### DIFF
--- a/src/common/proxy_wasm/ngx_proxy_wasm.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm.c
@@ -242,7 +242,12 @@ ngx_proxy_wasm_ctx_alloc(ngx_pool_t *pool)
         return NULL;
     }
 
-    pwctx->pool = pool;
+    pwctx->pool = ngx_create_pool(512, pool->log);
+    if (pwctx->pool == NULL) {
+        return NULL;
+    }
+
+    pwctx->parent_pool = pool;
 
     ngx_rbtree_init(&pwctx->host_props_tree, &pwctx->host_props_sentinel,
                     ngx_str_rbtree_insert_value);
@@ -294,20 +299,17 @@ ngx_proxy_wasm_ctx(ngx_uint_t *filter_ids, size_t nfilters,
                 return NULL;
             }
 
+            pool = pwctx->pool;
+            log = pwctx->log;
+
             switch (pwctx->isolation) {
             case NGX_PROXY_WASM_ISOLATION_NONE:
                 pwstore = filter->store;
-                pool = filter->pool;
-                log = pwctx->log;
                 break;
             case NGX_PROXY_WASM_ISOLATION_STREAM:
                 pwstore = &pwctx->store;
-                pool = pwstore->pool;
-                log = pwctx->log;
                 break;
             case NGX_PROXY_WASM_ISOLATION_FILTER:
-                pool = pwctx->pool;
-                log = pwctx->log;
                 break;
             default:
                 ngx_proxy_wasm_log_error(NGX_LOG_WASM_NYI, pwctx->log, 0,
@@ -399,6 +401,7 @@ ngx_proxy_wasm_ctx_destroy(ngx_proxy_wasm_ctx_t *pwctx)
 
     ngx_proxy_wasm_store_destroy(&pwctx->store);
 
+#if 0
     if (pwctx->authority.data) {
         ngx_pfree(pwctx->pool, pwctx->authority.data);
     }
@@ -442,10 +445,11 @@ ngx_proxy_wasm_ctx_destroy(ngx_proxy_wasm_ctx_t *pwctx)
     if (pwctx->response_status.data) {
         ngx_pfree(pwctx->pool, pwctx->response_status.data);
     }
+#endif
 
-    ngx_array_destroy(&pwctx->pwexecs);
+    ngx_destroy_pool(pwctx->pool);
 
-    ngx_pfree(pwctx->pool, pwctx);
+    ngx_pfree(pwctx->parent_pool, pwctx);
 }
 
 
--- a/src/wasm/wrt/ngx_wrt_wasmtime.c
+++ b/src/wasm/wrt/ngx_wrt_wasmtime.c
@@ -8,6 +8,10 @@
 #include <ngx_wasi.h>
 
 
+#define NGX_WRT_WASMTIME_STACK_ARGS  10
+#define NGX_WRT_WASMTIME_STACK_RETS  10
+
+
 typedef void (*wasmtime_config_set_int_pt)(wasm_config_t *config,
     uint64_t value);
 typedef void (*wasmtime_config_set_bool_pt)(wasm_config_t *config, bool value);
@@ -611,6 +615,9 @@ ngx_wasmtime_call(ngx_wrt_instance_t *instance, ngx_str_t *func_name,
     wasmtime_extern_t   item;
     wasmtime_func_t    *func;
     wasmtime_val_t     *wargs = NULL, *wrets = NULL;
+    wasmtime_val_t      swargs[NGX_WRT_WASMTIME_STACK_ARGS];
+    wasmtime_val_t      swrets[NGX_WRT_WASMTIME_STACK_RETS];
+
 
     if (!wasmtime_instance_export_get(instance->store->context,
                                       &instance->instance,
@@ -624,16 +631,28 @@ ngx_wasmtime_call(ngx_wrt_instance_t *instance, ngx_str_t *func_name,
 
     func = &item.of.func;
 
-    wargs = ngx_pcalloc(instance->pool, sizeof(wasmtime_val_t) * args->size);
-    if (wargs == NULL) {
-        goto done;
+    if (args->size <= NGX_WRT_WASMTIME_STACK_ARGS) {
+        wargs = &swargs[0];
+
+    } else {
+        wargs = ngx_pcalloc(instance->pool,
+                            sizeof(wasmtime_val_t) * args->size);
+        if (wargs == NULL) {
+            goto done;
+        }
     }
 
     ngx_wasm_valvec2wasmtime(wargs, args);
 
-    wrets = ngx_pcalloc(instance->pool, sizeof(wasmtime_val_t) * rets->size);
-    if (wrets == NULL) {
-        goto done;
+    if (rets->size <= NGX_WRT_WASMTIME_STACK_RETS) {
+        wrets = &swrets[0];
+
+    } else {
+        wrets = ngx_pcalloc(instance->pool,
+                            sizeof(wasmtime_val_t) * rets->size);
+        if (wrets == NULL) {
+            goto done;
+        }
     }
 
     err->res = wasmtime_func_call(instance->store->context,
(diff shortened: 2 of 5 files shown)


src/common/proxy_wasm/ngx_proxy_wasm.c
```diff
@@ -242,7 +242,8 @@ ngx_proxy_wasm_ctx_alloc(ngx_pool_t *pool)
         return NULL;
     }
 
-    pwctx->pool = pool;
+    pwctx->pool = ngx_create_pool(128, pool->log);
```
thibaultcha, line 245: The call can fail, let's catch the error; perhaps also make `128` a `#define`? It probably is a little low too, no?

src/wasm/wrt/ngx_wrt_wasmtime.c
```diff
@@ -611,6 +611,8 @@ ngx_wasmtime_call(ngx_wrt_instance_t *instance, ngx_str_t *func_name,
     wasmtime_extern_t   item;
     wasmtime_func_t    *func;
     wasmtime_val_t     *wargs = NULL, *wrets = NULL;
+    wasmtime_val_t      swargs[10], swrets[10];
```
thibaultcha, line 614: Let's also make 10 a #define?

src/wasm/wrt/ngx_wrt_wasmtime.c
```diff
@@ -624,16 +626,28 @@ ngx_wasmtime_call(ngx_wrt_instance_t *instance, ngx_str_t *func_name,
...
     func = &item.of.func;
 
-    wargs = ngx_pcalloc(instance->pool, sizeof(wasmtime_val_t) * args->size);
-    if (wargs == NULL) {
-        goto done;
+    if (args->size <= 10) {
+        wargs = &swargs[0];
+
+    } else {
+        wargs = ngx_pcalloc(instance->pool,
```
thibaultcha, line 633: It would be great to keep both of these cases covered via the `ngx_wasm_debug` module with a dummy host function interface perhaps.

src/common/proxy_wasm/ngx_proxy_wasm.c
thibaultcha: Another note too, I think we could first create the pool, then allocate `pwctx` inside of it? That way no need to keep track of the parent pool at all, and it could even disappear from this function's arguments.

thibaultcha (commented):
Very nice! And also good updates overall.

> Turns out that ngx_pfree, which was being used in the code to free args and rets, [does not free small allocations](https://github.com/nginx/nginx/blob/027b68168867514665751a65780f050952bc48ec/src/core/ngx_palloc.c#L277-L294). It only frees large allocations; small allocations are only freed when the pool itself is destroyed.

Hmm... Welp, that much I knew (that's also why the nopool patch exists), so my first thought upon reading your summary was to think that request pools are also destroyed... 🤔 So far its been the intention to attach everything to the request pool. So, checking on [ngx_http_proxy_wasm.c:361](https://github.com/Kong/ngx_wasm_module/blob/main/src/http/proxy_wasm/ngx_http_proxy_wasm.c#L361-L363) says that for main requests, we allocate the filter chain in `r->connection->pool`; I think that was a wrong choice, those pools might not get destroyed... Maybe using `r->pool` in all cases would have been a sufficient fix, but I do like better giving filter chains their own pools.

src/wasm/wrt/ngx_wrt_wasmtime.c
```diff
@@ -611,6 +611,8 @@ ngx_wasmtime_call(ngx_wrt_instance_t *instance, ngx_str_t *func_name,
     wasmtime_extern_t   item;
     wasmtime_func_t    *func;
     wasmtime_val_t     *wargs = NULL, *wrets = NULL;
```
thibaultcha, line 613: With the new code these also don't need to be initialized here anymore I think.

thibaultcha: Oh no, they still do.

src/common/proxy_wasm/ngx_proxy_wasm.c
thibaultcha: Yes, it is how it works. Maybe we can have `NGX_WASM_POOL_SIZE_S` and `NGX_WASM_POOL_SIZE_M` with `128` and `512`? Fine to hard-code `512` as well given we already do it, we can `#define` later.

hishamhm: > I think we could first create the pool, then allocate pwctx inside of it?

Can do, if you prefer this pattern. Personally, I always try to organize memory allocations as an ownership tree, that helps me track the lifetimes of allocations by looking at the structure definitions rather than by tracking the execution flow. Having the structure that owns the pool stored inside the pool itself makes the ownership relationship circular, which takes more attentive coding as it gets harder and harder to track all flows in your head as things grow, so it's a pattern I've always avoided. (I believe it's the kind of pattern you cannot even represent (easily?) in Rust, for example — I guess I've been "thinking in Rust" before Rust existed :sweat_smile: )

I wanted to avoid refactors to keep the PR small and the review easier, but if it were up to me, I'd prefer instead to make `ngx_proxy_wasm_ctx_alloc` and `ngx_proxy_wasm_ctx_destroy` more symmetrical not by removing the parent pool argument, but by passing it to the destroy function _as well_, making them more like `ngx_pcalloc` and `ngx_pfree` (which would then remove the `parent_pool` field from the struct).

That, of course, requires that the owner of the parent pool is reachable at destruction time. From a quick look at the code it seems that those alloc and destroy functions are called at pretty different places (one under `http/`, another under `common/`), so a change to make things more symmetrical would be more code surgery than I think should happen in this PR (though I still think that it could have potential benefits, at the very least to force a rethink/clarification of the pool lifetimes). I'm not a big fan of the `parent_pool` field either (I try to avoid "backlinks" to the owner structure whenever possible), but at least I like it better than the apparently-circular ownership. But I can go either way, no strong feelings here, just talking shop to share my thought process.

src/wasm/wrt/ngx_wrt_wasmtime.c
hishamhm: I took a look but I wasn't able to find my way right away... I'm assuming the idea is to add a new `ngx_command_t` from which to call `ngx_wavm_instance_call_func` (or `funcref`?) on a Wasm function with 11 args and 11 rets.

How do I get a `ngx_wavm_instance_t *` from the debug module? And how do I get an `ngx_wavm_func_t *` (or  `ngx_wavm_funcref_t *`?)

I'm assuming I'll need to add a .wat file in the test that declares the 11-arg and 11-ret function; that part I think I can figure out.

thibaultcha: Both the instance creation and the dummy host function interface will have to be coded into the debug module.

hishamhm: I managed to declare the host function interface and write the test case, but I tried to access the `wcf->vm` from `ngx_wasm_core` module to link the module, to avoid having to recreate the whole VM setup, and I think I ended up lost in module initialization ordering shenanigans. I think I might have ended up sinking more time trying to get this test case running than fixing the memory issue. If the lack of coverage is a blocker, I'm inclined to drop the `perf(wasmtime)` commit from this PR to get the `fix(proxy-wasm)` merged.

thibaultcha: We will do it later then.

hishamhm:
This Valgrind report looks legitimate:

```
==44506== 48 bytes in 1 blocks are definitely lost in loss record 5 of 33
==44506==    at 0x4848899: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==44506==    by 0x178092: ngx_alloc (ngx_alloc.c:22)
==44506==    by 0x1526A1: ngx_create_pool (ngx_palloc.c:21)
==44506==    by 0x2AAD1C: ngx_proxy_wasm_ctx_alloc (ngx_proxy_wasm.c:245)
==44506==    by 0x2BCA46: ngx_http_proxy_wasm_ctx (ngx_http_proxy_wasm.c:361)
==44506==    by 0x2B69E6: ngx_http_wasm_ffi_get_property (ngx_wasm_lua_ffi.c:218)
==44506==    by 0x4E94FE1: lj_vm_ffi_call (in /home/runner/work/ngx_wasm_module/ngx_wasm_module/work/buildroot/prefix/luajit/lib/libluajit-5.1.so.2.1.0)
==44506==    by 0x4EF44EF: lj_ccall_func (lj_ccall.c:1382)
==44506==    by 0x4F101C4: lj_cf_ffi_meta___call (lib_ffi.c:230)
==44506==    by 0x4E92B92: lj_BC_FUNCC (in /home/runner/work/ngx_wasm_module/ngx_wasm_module/work/buildroot/prefix/luajit/lib/libluajit-5.1.so.2.1.0)
==44506==    by 0x4EADD08: lua_pcall (lj_api.c:1145)
==44506==    by 0x238BA6: ngx_http_lua_do_call (ngx_http_lua_util.c:4159)
==44506==    by 0x250BCA: ngx_http_lua_init_worker_by_inline (ngx_http_lua_initworkerby.c:333)
==44506==    by 0x250A80: ngx_http_lua_init_worker (ngx_http_lua_initworkerby.c:298)
==44506==    by 0x17E385: ngx_single_process_cycle (ngx_process_cycle.c:299)
==44506==    by 0x15009C: main (nginx.c:383)
```

From what I could tell, it seems to happen because when calling an FFI function from `init_worker`, a`pwctx` gets created "on-demand" by `ngx_http_proxy_wasm_ctx`, but apparently the correspondent `ngx_proxy_wasm_ctx_destroy` never gets called (it is only called in a DONE step), so `pwctx->pool` never gets freed (destroying the `parent_pool` is not sufficient because pools have internal allocations).

I think before the lack of a `ngx_proxy_wasm_ctx_destroy` was harmless before because all it did was call various `ngx_pfree` on structures that ultimate were also freed when the whole pool was destroyed, but this might have uncovered an issue in the `pwctx` lifecycle management when using the FFI.

thibaultcha:
Patch with fix in #421. I will merge tomorrow; it will be a little tricky rebasing on my proxy_wasm.c refactor which also fixes some lingering memory issues with instances, most of the file has changed.

hishamhm:
Awesome, thanks! I've cherry-picked your commit. No further changes from my part unless you have further comments, so if you think it's ready, feel free to merge!

thibaultcha:
Merged!

PR #431, opened 2023-10-16
feat(proxy-wasm) host-managed property getters and setters

hishamhm:
This feature adds support for host-managed property getter and setter functions.

These feature is implemented in three stages:

* `ngx_proxy_wasm_properties.c`: in our implementation for proxy-wasm properties, we now allow the host to specify a function which gets called whenever proxy-wasm receives a `set_property` or `get_property` call matching the host-defined namespace. Those getter and setter functions are then responsible to return and store the values, and are able to produce any desired side-effects in the host.

* `ngx_wasm_lua_ffi.c`: We then expose this functionality in our Lua FFI interface. This is a lightweight wrapper, which accepts a C function pointer directly, and associates this function to the request context object.

* `lib/resty/wasmx/proxy_wasm.lua`: Finally, we expose a Lua API for setting a Lua functions as setters and getters. The interface is such that the host Lua program should set one setter function and/or getter function, and then within that function properly dispatch the behavior based on the received key. Internally, these functions use the fact that LuaJIT is able to convert a Lua function into a C function pointer. Due to lifetime management, those generated C callbacks are a limited resource in the LuaJIT VM, so we produce a single pair of these and then have them call the user-defined Lua function, which remains as a normal Lua closure. (See http://luajit.org/ext_ffi_semantics.html for more details on LuaJIT FFI C callback managment.)

To ensure that the lifetime of strings produced by the host continue valid by the time the getter callback finished running and the Wasm engine reads the properties, getter/setter-produced values continue to be stored in the rbtree for host properties that we already had for plain `get_property`/`set_property`, introduced in da820875ed7.

Since we need to store this data on the C side anyway, we also add an extra feature as an optimization: the concept of "const" host-properties. If the Lua callback returns a truthy third-argument, that signals that that value should be flagged as constant. Further calls to `get_property` during the same request will get the property value directly from the ngx_wasm_module rbtree "cache" without hitting the getter callback again. Test cases in the test suite demonstrate this feature.

In this PR, the host namespace continues to be statically defined at compile time via the NGX_WASM_HOST_PROPERTY_NAMESPACE define. The regular FFI functions for `set_property` and `get_property` introduced in 447ef130a9 also continue to work normally.

### DIFF
--- a/lib/resty/wasmx/proxy_wasm.lua
+++ b/lib/resty/wasmx/proxy_wasm.lua
@@ -8,9 +8,11 @@ local C = ffi.C
 local ngx = ngx
 local error = error
 local type = type
+local pcall = pcall
 local ffi_gc = ffi.gc
 local ffi_new = ffi.new
 local ffi_str = ffi.string
+local tostring = tostring
 local subsystem = ngx.config.subsystem
 local get_request = wasmx.get_request
 local get_err_ptr = wasmx.get_err_ptr
@@ -38,23 +40,32 @@ local _M = {
 
 if subsystem == "http" then
     ffi.cdef [[
-        int ngx_http_wasm_ffi_plan_new(ngx_wasm_vm_t *vm,
-                                       ngx_wasm_filter_t *filters,
-                                       size_t n_filters,
-                                       ngx_wasm_plan_t **out,
-                                       u_char *err, size_t *errlen);
-        int ngx_http_wasm_ffi_plan_free(ngx_wasm_plan_t *plan);
-        int ngx_http_wasm_ffi_plan_load(ngx_wasm_plan_t *plan);
-        int ngx_http_wasm_ffi_plan_attach(ngx_http_request_t *r,
-                                          ngx_wasm_plan_t *plan,
-                                          unsigned int isolation);
-        int ngx_http_wasm_ffi_start(ngx_http_request_t *r);
-        int ngx_http_wasm_ffi_set_property(ngx_http_request_t *r,
-                                           ngx_str_t *key,
-                                           ngx_str_t *value);
-        int ngx_http_wasm_ffi_get_property(ngx_http_request_t *r,
-                                           ngx_str_t *key,
-                                           ngx_str_t *out);
+        ngx_int_t ngx_http_wasm_ffi_plan_new(ngx_wasm_vm_t *vm,
+                                             ngx_wasm_filter_t *filters,
+                                             size_t n_filters,
+                                             ngx_wasm_plan_t **out,
+                                             u_char *err, size_t *errlen);
+        ngx_int_t ngx_http_wasm_ffi_plan_free(ngx_wasm_plan_t *plan);
+        ngx_int_t ngx_http_wasm_ffi_plan_load(ngx_wasm_plan_t *plan);
+        ngx_int_t ngx_http_wasm_ffi_plan_attach(ngx_http_request_t *r,
+                                                ngx_wasm_plan_t *plan,
+                                                unsigned int isolation);
+        ngx_int_t ngx_http_wasm_ffi_start(ngx_http_request_t *r);
+        ngx_int_t ngx_http_wasm_ffi_set_property(ngx_http_request_t *r,
+                                                 ngx_str_t *key,
+                                                 ngx_str_t *value);
+        ngx_int_t ngx_http_wasm_ffi_get_property(ngx_http_request_t *r,
+                                                 ngx_str_t *key,
+                                                 ngx_str_t *out);
+        ngx_int_t ngx_http_wasm_ffi_set_host_property(ngx_http_request_t *r,
+                                                      ngx_str_t *key,
+                                                      ngx_str_t *value,
+                                                      unsigned is_const,
+                                                      unsigned retrieve);
+        ngx_int_t ngx_http_wasm_ffi_set_property_setter(ngx_http_request_t *r,
+                                                        ngx_wasm_host_prop_fn_t fn);
+        ngx_int_t ngx_http_wasm_ffi_set_property_getter(ngx_http_request_t *r,
+                                                        ngx_wasm_host_prop_fn_t fn);
     ]]
 
 else
@@ -65,6 +76,86 @@ end
 
 local ffi_ops_type = ffi.typeof("ngx_wasm_plan_t *")
 local ffi_pops_type = ffi.typeof("ngx_wasm_plan_t *[1]")
+
+
+local lua_prop_setter
+local lua_prop_getter
+local c_prop_setter
+local c_prop_getter
+
+
+do
+    -- LuaJIT C callbacks are a limited resource:
+    -- We can't create one of these on each request,
+    -- so we prepare these wrappers, which in turn call a Lua function.
+
+    local function store_c_value(r, ckey, cvalue, lvalue, is_const)
+        if lvalue ~= nil then
+            local value = tostring(lvalue)
+            cvalue.data = value
+            cvalue.len = #value
+
+        else
+            cvalue.data = nil
+            cvalue.len = 0
+        end
+
+        return C.ngx_http_wasm_ffi_set_host_property(r, ckey, cvalue,
+                                                     is_const and 1 or 0,
+                                                     1)
+    end
+
+    c_prop_setter = ffi.cast("ngx_wasm_host_prop_fn_t", function(r, ckey,
+                                                                 cvalue)
+        local lkey = ffi_str(ckey.data, ckey.len)
+        local lval
+        if cvalue.data ~= nil then
+            lval = ffi_str(cvalue.data, cvalue.len)
+        end
+
+        local pok, ok, lvalue, is_const = pcall(lua_prop_setter, lkey, lval)
+        if not pok then
+            ngx.log(ngx.ERR, "error setting property from Lua: ", ok)
+            return FFI_ERROR
+        end
+
+        if not ok then
+            local err = lvalue or "unknown error"
+            ngx.log(ngx.ERR, "error setting property: ", err)
+            return FFI_ERROR
+        end
+
+        return store_c_value(r, ckey, cvalue, lvalue, is_const)
+    end)
+
+
+    c_prop_getter = ffi.cast("ngx_wasm_host_prop_fn_t", function(r, ckey,
+                                                                 cvalue)
+        local lkey = ffi_str(ckey.data, ckey.len)
+
+        local pok, ok, lvalue, is_const = pcall(lua_prop_getter, lkey)
+        if not pok then
+            ngx.log(ngx.ERR, "error getting property from Lua: ", ok)
+            return FFI_ERROR
+        end
+
+        if not ok then
+            if lvalue then
+                ngx.log(ngx.ERR, "error setting property: ", lvalue)
+                return FFI_ERROR
+            end
+
+            return FFI_DECLINED
+        end
+
+        local rc = store_c_value(r, ckey, cvalue, lvalue, is_const)
+        if rc == FFI_OK and lvalue == nil then
+            return FFI_DECLINED
+        end
+
+        return rc
+    end)
+end
 
 
 function _M.new(filters)
@@ -229,7 +320,7 @@ function _M.set_property(key, value)
         error("key must be a string", 2)
     end
 
-    if type(value) ~= "string" then
+    if value ~= nil and type(value) ~= "string" then
         error("value must be a string", 2)
     end
 
@@ -239,7 +330,7 @@ function _M.set_property(key, value)
     end
 
     local ckey = ffi_new("ngx_str_t", { data = key, len = #key })
-    local cvalue = ffi_new("ngx_str_t", { data = value, len = #value })
+    local cvalue = ffi_new("ngx_str_t", { data = value, len = value and #value or 0 })
 
     local rc = C.ngx_http_wasm_ffi_set_property(r, ckey, cvalue)
     if rc == FFI_ERROR then
@@ -277,6 +368,75 @@ function _M.get_property(key)
     end
 
     return ffi_str(cvalue.data, cvalue.len)
+end
+
+
+---
+-- Define a setter function for host-managed properties.
+--
+-- @param setter The setter function is the handler for updating properties.
+-- Its type signature is `function(string, string): boolean, string`,
+-- where the inputs are the property key and value and the results are:
+-- * `truthy` - success setting the property
+-- * `truthy, value` - success, cache property value (useful if hosts
+--   sets only a setter but not a getter)
+-- * `truthy, value, truthy` - success, constant property value: further
+--   requests to the same property during a request are cached by
+--   ngx_wasm_module and do not invoke the Lua getter.
+-- * `falsy, err` - failure, and an optional error message
+-- @return true on success setting the getter, or nil and an error message.
+function _M.set_property_setter(setter)
+    if type(setter) ~= "function" then
+        error("setter must be a function", 2)
+    end
+
+    local r = get_request()
+    if not r then
+        error("no request found", 2)
+    end
+
+    lua_prop_setter = setter
+
+    local rc = C.ngx_http_wasm_ffi_set_property_setter(r, c_prop_setter)
+    if rc == FFI_OK then
+        return true
+    end
+
+    return nil, "could not set property setter"
+end
+
+
+---
+-- Define a getter function for host-managed properties.
+--
+-- @param getter The getter function is the handler for resolving properties.
+-- Its type signature is `function(string): boolean, string, boolean`,
+-- where the input is the property key and the results are:
+-- * `truthy, value` - success, property value
+-- * `truthy, value, truthy` - success, constant property value: further
+--   requests to the same property during a request are cached by
+--   ngx_wasm_module and do not invoke the Lua getter.
+-- * `falsy` - property not found
+-- * `falsy, err` - failure, error message
+-- @return true on success setting the getter, or nil and an error message.
+function _M.set_property_getter(getter)
+    if type(getter) ~= "function" then
+        error("getter must be a function", 2)
+    end
+
+    local r = get_request()
+    if not r then
+        error("no request found", 2)
+    end
+
+    lua_prop_getter = getter
+
+    local rc = C.ngx_http_wasm_ffi_set_property_getter(r, c_prop_getter)
+    if rc == FFI_OK then
+        return true
+    end
+
+    return nil, "could not set property getter"
 end
 
 
--- a/src/common/proxy_wasm/ngx_proxy_wasm_properties.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm_properties.c
@@ -53,6 +53,8 @@ static size_t       host_prefix_len;
 typedef struct {
     ngx_str_node_t  sn;
     ngx_str_t       value;
+    unsigned        is_const:1;
+    unsigned        negative_cache:1;
 } host_props_node_t;
 
 
@@ -876,25 +878,58 @@ ngx_proxy_wasm_properties_get_host(ngx_proxy_wasm_ctx_t *pwctx,
 
     hpn = (host_props_node_t *)
               ngx_str_rbtree_lookup(&pwctx->host_props_tree, path, hash);
-
-    if (hpn) {
-        value->data = hpn->value.data;
-        value->len = hpn->value.len;
-
-        return NGX_OK;
+    if (!hpn) {
+        return NGX_DECLINED;
     }
 
-    return NGX_DECLINED;
+    /* stored value is not const: need to call getter again */
+    if (!hpn->is_const && pwctx->host_prop_getter) {
+        return NGX_DECLINED;
+    }
+
+    value->data = hpn->value.data;
+    value->len = hpn->value.len;
+
+    if (hpn->negative_cache) {
+        return NGX_BUSY;
+    }
+
+    return NGX_OK;
 }
 
 
-static ngx_int_t
+static ngx_inline ngx_int_t
+set_hpn_value(host_props_node_t *hpn, ngx_str_t *value, unsigned is_const,
+    ngx_pool_t *pool)
+{
+    unsigned char  *new_data;
+
+    if (value->data == NULL) {
+        new_data = NULL;
+        hpn->negative_cache = 1;
+
+    } else {
+        new_data = ngx_pstrdup(pool, value);
+        if (new_data == NULL) {
+            return NGX_ERROR;
+        }
+    }
+
+    hpn->is_const = is_const;
+    hpn->value.len = value->len;
+    hpn->value.data = new_data;
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
 ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
-    ngx_str_t *path, ngx_str_t *value)
+    ngx_str_t *path, ngx_str_t *value, unsigned is_const, unsigned retrieve)
 {
     uint32_t                  hash;
-    unsigned                  new_entry = 1;
     host_props_node_t        *hpn;
+    ngx_int_t                 rc;
 #ifdef NGX_WASM_HTTP
     ngx_http_wasm_req_ctx_t  *rctx = pwctx->data;
 
@@ -910,17 +945,27 @@ ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
     hpn = (host_props_node_t *)
               ngx_str_rbtree_lookup(&pwctx->host_props_tree, path, hash);
 
-    if (value->data == NULL) {
-        if (hpn) {
-            ngx_rbtree_delete(&pwctx->host_props_tree, &hpn->sn.node);
+    if (hpn) {
+        if (hpn->is_const) {
+            return NGX_DECLINED;
         }
 
-        return NGX_OK;
-    }
-
-    if (hpn) {
         ngx_pfree(pwctx->pool, hpn->value.data);
-        new_entry = 0;
+
+        if (value->data == NULL && !is_const) {
+            ngx_rbtree_delete(&pwctx->host_props_tree, &hpn->sn.node);
+
+            return NGX_OK;
+        }
+
+        rc = set_hpn_value(hpn, value, is_const, pwctx->pool);
+        if (rc != NGX_OK) {
+            return rc;
+        }
+
+    } else if (value->data == NULL && !is_const) {
+        /* no need to store a non-const NULL */
+        return NGX_OK;
 
     } else {
         hpn = ngx_pcalloc(pwctx->pool, sizeof(host_props_node_t) + path->len);
@@ -932,16 +977,17 @@ ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
         hpn->sn.str.len = path->len;
         hpn->sn.str.data = (u_char *) hpn + sizeof(host_props_node_t);
         ngx_memcpy(hpn->sn.str.data, path->data, path->len);
-    }
 
-    hpn->value.len = value->len;
-    hpn->value.data = ngx_pstrdup(pwctx->pool, value);
-    if (hpn->value.data == NULL) {
-        return NGX_ERROR;
-    }
+        rc = set_hpn_value(hpn, value, is_const, pwctx->pool);
+        if (rc != NGX_OK) {
+            return rc;
+        }
 
-    if (new_entry) {
         ngx_rbtree_insert(&pwctx->host_props_tree, &hpn->sn.node);
+    }
+
+    if (retrieve) {
+        value->data = hpn->value.data;
     }
 
     return NGX_OK;
@@ -969,6 +1015,7 @@ ngx_proxy_wasm_properties_get(ngx_proxy_wasm_ctx_t *pwctx,
     ngx_str_t           p = { path->len, NULL };
     ngx_uint_t          key;
     pwm2ngx_mapping_t  *m;
+    ngx_int_t           rc;
 
     p.data = replace_nulls_by_dots(path, dotted_path_buf);
     key = ngx_hash_key(p.data, p.len);
@@ -996,7 +1043,22 @@ ngx_proxy_wasm_properties_get(ngx_proxy_wasm_ctx_t *pwctx,
         && ngx_memcmp(p.data, host_prefix, host_prefix_len) == 0)
     {
         /* host variable */
-        return ngx_proxy_wasm_properties_get_host(pwctx, &p, value);
+
+        /* even if there is a getter, try reading const value first */
+        rc = ngx_proxy_wasm_properties_get_host(pwctx, &p, value);
+        if (rc == NGX_OK) {
+            return NGX_OK;
+
+        } else if (rc == NGX_BUSY) {
+            return NGX_DECLINED;
+        }
+
+        if (pwctx->host_prop_getter) {
+            return pwctx->host_prop_getter(pwctx->host_prop_getter_data,
+                                           &p, value);
+        }
+
+        return rc;
     }
 
     return NGX_DECLINED;
@@ -1032,8 +1094,57 @@ ngx_proxy_wasm_properties_set(ngx_proxy_wasm_ctx_t *pwctx,
         && ngx_memcmp(p.data, host_prefix, host_prefix_len) == 0)
     {
         /* host variable */
-        return ngx_proxy_wasm_properties_set_host(pwctx, &p, value);
+        if (pwctx->host_prop_setter) {
+            return pwctx->host_prop_setter(pwctx->host_prop_setter_data,
+                                           &p, value);
+        }
+
+        return ngx_proxy_wasm_properties_set_host(pwctx, &p, value, 0, 0);
     }
 
     return NGX_DECLINED;
+}
+
+
+ngx_int_t
+ngx_proxy_wasm_properties_set_host_prop_setter(ngx_proxy_wasm_ctx_t *pwctx,
+    ngx_wasm_host_prop_fn_t fn, void *data)
+{
+#ifdef NGX_WASM_HTTP
+    ngx_http_wasm_req_ctx_t  *rctx = pwctx->data;
+
+    if (rctx == NULL || rctx->fake_request) {
+        ngx_wavm_log_error(NGX_LOG_ERR, pwctx->log, NULL,
+                           "cannot set host properties setter "
+                           "outside of a request");
+        return NGX_ERROR;
+    }
+#endif
+
+    pwctx->host_prop_setter = fn;
+    pwctx->host_prop_setter_data = data;
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
+ngx_proxy_wasm_properties_set_host_prop_getter(ngx_proxy_wasm_ctx_t *pwctx,
+    ngx_wasm_host_prop_fn_t fn, void *data)
+{
+#ifdef NGX_WASM_HTTP
+    ngx_http_wasm_req_ctx_t  *rctx = pwctx->data;
+
+    if (rctx == NULL || rctx->fake_request) {
+        ngx_wavm_log_error(NGX_LOG_ERR, pwctx->log, NULL,
+                           "cannot set host properties getter "
+                           "outside of a request");
+        return NGX_ERROR;
+    }
+#endif
+
+    pwctx->host_prop_getter = fn;
+    pwctx->host_prop_getter_data = data;
+
+    return NGX_OK;
 }
--- a/t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
+++ b/t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
@@ -1,0 +1,597 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+use strict;
+use lib '.';
+use t::TestWasm::Lua;
+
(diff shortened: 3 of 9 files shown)


lib/resty/wasmx/proxy_wasm.lua
```diff
@@ -67,6 +76,78 @@ local ffi_ops_type = ffi.typeof("ngx_wasm_plan_t *")
...
+
+        return C.ngx_http_wasm_ffi_set_host_property(r, ckey, cvalue,
+                                                     is_const and 1 or 0,
+                                                     1)
+    end
+
+    c_prop_setter = ffi.cast("ngx_wasm_host_prop_fn_t", function(r, ckey,
+                                                                 cvalue)
+        local lkey = ffi_str(ckey.data, ckey.len)
+        local lval = ffi_str(cvalue.data, cvalue.len)
```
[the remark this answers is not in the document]
hishamhm: I've added code to handle this case, and to make the behavior of the `nil` case consistent with const properties as well. I implemented this by adding a negative-cache flag for const-properties on the C side: that was cleaner than trying to encode that case implicitly in the `value`, since empty-strings are 0-length `ngx_str_t` entries as well.

thibaultcha:
Not able to make a full review at this time but at merge time, for sure the 2nd commit should be squashed, at least.

hishamhm:
> Not able to make a full review at this time but at merge time, for sure the 2nd commit should be squashed, at least.

Yeah, for sure, I only kept it separate for readability during review and it helps to scope what that extra code is for.

thibaultcha (commented):
Quick review + have not looked at the tests either.

lib/resty/wasmx/proxy_wasm.lua
```diff
@@ -38,23 +38,32 @@ local _M = {
...
-                                          ngx_wasm_plan_t *plan,
-                                          unsigned int isolation);
-        int ngx_http_wasm_ffi_start(ngx_http_request_t *r);
-        int ngx_http_wasm_ffi_set_property(ngx_http_request_t *r,
-                                           ngx_str_t *key,
-                                           ngx_str_t *value);
-        int ngx_http_wasm_ffi_get_property(ngx_http_request_t *r,
-                                           ngx_str_t *key,
-                                           ngx_str_t *out);
+        ngx_int_t ngx_http_wasm_ffi_plan_new(ngx_wasm_vm_t *vm,
```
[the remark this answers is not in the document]
hishamhm: There is: `sizeof(int) == 4`, `sizeof(ngx_int_t) == 8`. Using `int` in a function signature that required `ngx_int_t` caused me signal extension errors in the C side when propagating negative error codes. So since I needed to use the correct type for the function I was hitting, I figured I'd fix it as well for all the other ones, lest a similar bug might be lurking somewhere.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -911,16 +918,19 @@ ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
...
+            return NGX_DECLINED;
         }
 
-        return NGX_OK;
-    }
-
-    if (hpn) {
         ngx_pfree(pwctx->pool, hpn->value.data);
+
+        if (value->data == NULL && !is_const) {
```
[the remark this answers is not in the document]
hishamhm: It is, because:
* for non-const, setting a `NULL` value means removing an entry from the rbtree;
* for const, setting a `NULL` value means storing a negative-cache entry, so that further gets to the key don't hit the Lua getter
 
That is, preserving the "const cache" behavior consistently for all values returned by the Lua setter, including `return true, null, true`  meaning "success, value is `null`, cache this value".

lib/resty/wasmx/proxy_wasm.lua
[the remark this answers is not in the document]
thibaultcha: Right but... int being signed, which return value exactly? 🤔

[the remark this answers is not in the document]
hishamhm: It happened to `rc` values other than `NGX_OK`, which are negative, such as `NGX_DECLINED` (-5). IIRC it happened when passing that value through the LuaJIT callback typedef. That signed negative 32-bit number would auto-expand into a 64-bit value without sign-extension happening, so I was seeing a high positive number (typical 4-billion-something, probably 2^32-5) in the C side where -5 was expected, which fixed itself when I changed the FFI definition on the Lua side from `int` to the correct definition of `ngx_int_t` (`intptr_t`, which is 64-bit and predefined by LuaJIT).

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -878,23 +880,28 @@ ngx_proxy_wasm_properties_get_host(ngx_proxy_wasm_ctx_t *pwctx,
     hpn = (host_props_node_t *)
               ngx_str_rbtree_lookup(&pwctx->host_props_tree, path, hash);
 
-    if (hpn) {
+    if (hpn && (hpn->is_const || pwctx->host_prop_getter == NULL)) {
         value->data = hpn->value.data;
         value->len = hpn->value.len;
 
+        if (hpn->negative_cache) {
+            *negative_cache = 1;
```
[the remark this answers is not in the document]
hishamhm: Can do! Generally I'd avoid inventing such purpose-specific meanings to API-provided constants, but looks like it's a common pattern in Nginx code, and it will streamline it anyway in a way which looks clear enough. :+1:

lib/resty/wasmx/proxy_wasm.lua
```diff
@@ -67,6 +76,88 @@ local ffi_ops_type = ffi.typeof("ngx_wasm_plan_t *")
...
+
+
+do
+    -- LuaJIT C callbacks are a limited resource:
+    -- We can't create one of these on each request,
+    -- so we prepare these wrappers, which in turn call a Lua function.
+
+    local function store_c_value(r, ckey, cvalue, lvalue, is_const)
+        if lvalue ~= nil then
+            local value = tostring(lvalue)
```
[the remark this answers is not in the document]
hishamhm: I'm not used to it anymore because Teal caches the stdlib for me :laughing:

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -935,16 +945,29 @@ ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
         ngx_memcpy(hpn->sn.str.data, path->data, path->len);
     }
 
-    hpn->value.len = value->len;
-    hpn->value.data = ngx_pstrdup(pwctx->pool, value);
-    if (hpn->value.data == NULL) {
-        return NGX_ERROR;
+    if (value->data == NULL) {
```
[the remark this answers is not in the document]
hishamhm: > Can this branch be taken when hpn == NULL?

No, because the `else` branch in the test above creates a new `hpn` if it doesn't exist. The flow ended up a bit tricky because of the `is_const` and `!is_const` cases.

The iteration you reviewed also contained one inefficiency (if you tried to `set_property` to `NULL` and there wasn't one there before, it would store one non-const entry).

Fixing that added yet another `if` case, so I ended up reworking the whole thing to make it look more linear, which made the code a bit longer but hopefully easier to follow and just as efficient.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -935,16 +945,29 @@ ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
...
+    hpn->is_const = is_const;
+    hpn->value.len = value->len;
+    hpn->value.data = new_data;
+
     if (new_entry) {
         ngx_rbtree_insert(&pwctx->host_props_tree, &hpn->sn.node);
     }
 
+    if (retrieve) {
+        value->data = hpn->value.data;
```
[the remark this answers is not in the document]
hishamhm: To ensure we get our string lifetimes correct without doing a double roundtrip between Lua and C:

* When Wasm code calls `get_property`, we want to trigger a Lua getter.
* That Lua getter produces a Lua string, which is a pointer to Lua VM-managed data.
* The Lua getter finishes running, and then the WasmX module needs to copy that string over to the Wasm linear memory.
* However, there's a chance that when the Lua getter finishes running, the Lua GC triggers and drops the string, because we were not holding a Lua reference to it, but rather a raw C pointer (the `data` field of `ngx_str_t`).
* For this reason, we need to copy the Lua string to C memory to be held by the WasmX module. That's why the getter calls `store_c_value`. 
* But when we store the Lua string as a C value, we want our `ngx_str_t` to contain a pointer to the C-land copy of the string, not the original Lua-managed string. Hence, we ask `set_host_property` to set the host property and also `retrieve` the updated pointer.

We could also "set" then "get", but the default behavior of the Lua-accessible "get" is to call the getter if set, so we'd need something like a "rawget" of sorts to avoid a loop, and even if we had that, it would be two FFI call roundtrips, so I figured the `retrieve` business would be the least-bad option.

Alternatively, we could simplify the API by making `ngx_proxy_wasm_properties_set` always "retrieve" the copied `value` (that change would be harmless to both places where this function is currently used, namely, `ffi` and `hfuncs`).

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -878,23 +880,28 @@ ngx_proxy_wasm_properties_get_host(ngx_proxy_wasm_ctx_t *pwctx,
     hpn = (host_props_node_t *)
               ngx_str_rbtree_lookup(&pwctx->host_props_tree, path, hash);
 
-    if (hpn) {
+    if (hpn && (hpn->is_const || pwctx->host_prop_getter == NULL)) {
```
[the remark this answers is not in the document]
hishamhm: Should look nicer now!

lib/resty/wasmx/proxy_wasm.lua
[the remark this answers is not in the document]
thibaultcha: I do not understand how that is possible. We already have test cases with negative return values such as `NGX_ERROR` (-1) and `NGX_ABORT` (-6), and `int` should already be signed and able to carry these error values. Something else is probably at play here?

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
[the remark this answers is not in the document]
thibaultcha: Ok I see, thanks!

lib/resty/wasmx/proxy_wasm.lua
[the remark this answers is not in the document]
hishamhm: > Something else is probably at play here?

I believe it's LuaJIT's autoconversion of Lua closure into C function pointers. This is new.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -877,25 +879,58 @@ ngx_proxy_wasm_properties_get_host(ngx_proxy_wasm_ctx_t *pwctx,
...
-static ngx_int_t
+static ngx_inline ngx_int_t
+set_hpn_value(host_props_node_t *hpn, ngx_str_t *value, unsigned is_const,
+    ngx_pool_t *pool)
+{
+    unsigned char  *new_data;
+
+    if (value->data == NULL) {
+        new_data = NULL;
+        hpn->negative_cache = 1;
```
thibaultcha, line 910: Isn't `negative_cache` supposed to only be for consts?

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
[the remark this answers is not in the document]
thibaultcha: So should `store_c_value` when invoked from `c_prop_setter` not have this `retrieve` argument set? Or rather, not asking for a change, only double-checking that in that case, it is unnecessary. I think that is what confused me the most, why we'd need to "get" the value when we are just setting it.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -911,17 +946,27 @@ ngx_proxy_wasm_properties_set_host(ngx_proxy_wasm_ctx_t *pwctx,
...
+        if (rc != NGX_OK) {
+            return rc;
+        }
 
-    if (hpn) {
-        ngx_pfree(pwctx->pool, hpn->value.data);
-        new_entry = 0;
+    } else if (value->data == NULL && !is_const) {
+        /* no need to store a non-const NULL */
+        return NGX_OK;
```
thibaultcha, line 969: Can we have this early-exit branch on top as a standalone `if {}`? Overall I find this entire flow **extremely** difficult to read and follow, especially given the APIs aren't symmetrical anymore and `set_host_property` is now an entrypoint (instead of `set_property` only), so I am trying to find ways to make it as clear as possible. Will probably update a lot of it during merge as well...

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -877,25 +879,58 @@ ngx_proxy_wasm_properties_get_host(ngx_proxy_wasm_ctx_t *pwctx,
 
     hpn = (host_props_node_t *)
               ngx_str_rbtree_lookup(&pwctx->host_props_tree, path, hash);
+    if (!hpn) {
```
thibaultcha, line 882:
```suggestion
    if (hpn == NULL) {
```

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -877,25 +879,58 @@ ngx_proxy_wasm_properties_get_host(ngx_proxy_wasm_ctx_t *pwctx,
...
+    /* stored value is not const: need to call getter again */
+    if (!hpn->is_const && pwctx->host_prop_getter) {
+        return NGX_DECLINED;
+    }
 
-        return NGX_OK;
+    value->data = hpn->value.data;
+    value->len = hpn->value.len;
+
+    if (hpn->negative_cache) {
```
thibaultcha, line 894: Should this be hoisted up above the previous 2 lines or do they have to run in this case? I keep forgetting what the negative cache is about, maybe some comments would help.

thibaultcha (commented):
Not easy to review post eye-surgery but I gave it another round

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
thibaultcha: It does seem to be for consts only; maybe we only set this bit when `is_const` is enabled? This way it'd be easier to trace-out what this bit tries to relate?

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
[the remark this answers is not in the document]
hishamhm: You're right, `retrieve` doesn't need to be set in the "setter" case.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
hishamhm: yes, but for non-consts this function is never called.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
hishamhm: I renamed `negative_cache` to `const_null`

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
hishamhm: I reworked the flow of the function again!

hishamhm:
@thibaultcha That LuaJIT FFI callback issue is proving to be way more annoying than anticipated. :( I guess I celebrated victory too soon; we had _one_ lucky full run on the DB-less tests, but Postgres-based integration tests keep failing (and in a not fully consistent way, as is typical of code that works or not depending on FFI heuristics): https://github.com/Kong/kong/pull/11856

I'm afraid we'll have to rethink the communication strategy away from the slick-but-unreliable FFI callbacks and into the boring-but-faithful Lua/C API approach used by the Lua bridge. Do we have access to the `lua_State` running in the OpenResty worker? (That is, if we use the Lua bridge from WasmX, will we have access to Kong Gateway's global state?)

thibaultcha:
@hishamhm Yes, the WasmX Lua bridge provides access to the OpenResty `lua_State` with `lctx->L`, which we get from `ngx_http_lua_get_lua_vm`.

lib/resty/wasmx/proxy_wasm.lua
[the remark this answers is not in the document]
thibaultcha: If we don't need this anymore I'd be in favor of reverting it to keep the FFI as simple as possible.

hishamhm:
@thibaultcha Reverted back to the original implementation, with the original implementation's reviews squashed into the commit. Should be ready to go!

lib/resty/wasmx/proxy_wasm.lua
```diff
@@ -38,23 +40,32 @@ local _M = {
...
+                                                 ngx_str_t *out);
+        ngx_int_t ngx_http_wasm_ffi_set_host_property(ngx_http_request_t *r,
+                                                      ngx_str_t *key,
+                                                      ngx_str_t *value,
+                                                      unsigned is_const,
+                                                      unsigned retrieve);
+        ngx_int_t ngx_http_wasm_ffi_set_property_setter(ngx_http_request_t *r,
+                                                        ngx_wasm_host_prop_fn_t fn);
+        ngx_int_t ngx_http_wasm_ffi_set_property_getter(ngx_http_request_t *r,
+                                                        ngx_wasm_host_prop_fn_t fn);
```
hishamhm, line 68: @thibaultcha I still think we should keep the `ngx_int_t` changes because they fix the observable `sizeof` of the return values of these FFI functions to match what's really happening in C... even if most of them are not giving us trouble right now, that might bite us in the future is we keep the discrepancy around.

(As for code aesthetics, I'd be down with moving that string to an unindented top-level thing like
```lua

local ffi_http_cdefs = [[
ngx_int_t ngx_http_wasm_ffi_plan_new(ngx_wasm_vm_t *vm,
...
]]

-- ...


if subsystem == "http" then
    ffi.cdef(ffi_http_cdefs)
```
)

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
thibaultcha: It seems like many changes from the early reviews are gone?

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
thibaultcha: This as well (already updated locally, but then started noticing more; is the PR currently lacking the reviews changes?)

thibaultcha: Or did those changes never made it? Since I already started updating the PR locally please refrain from making more updates, unless it is indeed missing a large set of review changes...

hishamhm: ohh I might have reset the branch too far back in time?... I'm pretty sure I renamed the `negative_cache` stuff. Let me look it up.

hishamhm: @thibaultcha Yes, something on my end went out of sync (my local checkout shows `origin` for this branch as containing the updated code, which it clearly doesn't).

To simplify things and not get in the way of your merge, I've pushed the commit with the applied reviews in a separate branch: https://github.com/Kong/ngx_wasm_module/tree/feat/ffi-properties-2

I've also triggered a CI run: https://github.com/Kong/ngx_wasm_module/actions/runs/6934996957

Edit: for ease of merge I also pushed a branch with the code reviews in a separate commit, which I found locally: 
[feat/ffi-properties-code-review](https://github.com/Kong/ngx_wasm_module/tree/feat/ffi-properties-code-review)

lib/resty/wasmx/proxy_wasm.lua
thibaultcha: So what I encounter because of this `int` to `ngx_int_t` change is that return values (`rc`) from FFI calls such as `C.ngx_http_wasm_ffi_set_host_property` are returned as `cdata` types instead of `number`, which is quite the undesired side-effect imho, especially given `int` should already cover all necessary return codes, just like in the core OpenResty FFI APIs.

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
```diff
@@ -0,0 +1,597 @@
...
+[
+    qr/\[error\] .*? cannot set host properties getter outside of a request/,
+    qr/\[error\] .*? init_worker_by_lua(\(nginx\.conf:\d+\))?:\d+: could not set property getter/,
+]
+--- no_error_log
+[crit]
+
+
+
+=== TEST 2: set_property_getter() - setting host properties getter works on rewrite_by_lua
```
thibaultcha, line 54: I would like to understand what happens in this test because it seems wrong, but I don't. We ask for a log of `wasmx.my_property` in `on_response_body` (which is called twice), and we get:

```
2023/11/21 17:14:07 [debug] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] filter 1/1 resuming "on_response_body" step in "body_filter" phase                                                                         
2023/11/21 17:14:07 [info] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] [hostcalls] on_response_body, 3 bytes, eof: false, client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"      │
2023/11/21 17:14:07 [info] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] [hostcalls] testing in "ResponseBody", client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"
...                                                 
2023/11/21 17:14:07 [info] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] wasmx.my_property: my_value, client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"                            
...
2023/11/21 17:14:07 [debug] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] filter 1/1 resuming "on_response_body" step in "body_filter" phase                                                                         
2023/11/21 17:14:07 [info] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] [hostcalls] on_response_body, 0 bytes, eof: true, client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"
2023/11/21 17:14:07 [info] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] [hostcalls] testing in "ResponseBody", client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"
...                                                                                                                                                                                             
2023/11/21 17:14:07 [info] 1992222#0: *2 [proxy-wasm]["hostcalls" #1] property not found: wasmx.my_property, client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"
```

I was trying to understand why this branch is being taken in `set_hpn_value` in this particular test case:

```c
    if (hpn->const_null) {
        return NGX_BUSY;
    }
```

(I did the rename to `const_null` myself because it still is not in the latest branch with code review).

Originally my goal was to find which tests were using this branch and I saw TEST 10, but this TEST 2 is concerning to me.

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
```diff
@@ -0,0 +1,597 @@
...
+            assert(proxy_wasm.start())
+            ngx.say("ok")
+        }
+    }
+--- response_body
+ok
+--- error_log eval
+[
+    qr/\[info\] .*? \[hostcalls\] on_response_body/,
+    qr/\[info\] .*? wasmx.my_property: my_value/,
```
thibaultcha, line 94: Recently I have stopped relying on `error_log` for my test cases because in my experience they result in false-positives, like this test atm. If not false-positive, they also fail to encompass larger edge-cases such as this one in my above comment. Furthermore, there is no way of knowing if `on_response_body` (the 1st regex) does happen before the 2nd one, or not. For rock-solid, tight test cases, I advise using `grep_error_log_out` at all times, extracting the necessary patterns such that even `wasmx.my_property: not found` string would also come up and make this test properly fail (at least in its current test).

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
```diff
@@ -0,0 +1,597 @@
...
+        }
+
+        _G.c_plan = assert(proxy_wasm.new(filters))
+        assert(proxy_wasm.load(_G.c_plan))
+    }
+--- config
+    location /t {
+        proxy_wasm hostcalls 'on=response_body \
+                              test=/t/log/property \
+                              name=wasmx.my_property';
```
thibaultcha, line 70: Here it seems we can get rid of this filter and configure the Lua-attached one to do the same thing instead?

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
thibaultcha: Apparently something that I changed caused this behavior change, trying to find it. The original branch correctly shows `wasmx.my_property: my_value` for both `on_response_body` invocations, but since the test is not robust enough (see other comment), it did not catch the change when I made it. Now I can't find which change it was quite yet, since I committed it in a WIP commit.

thibaultcha: Something seems amiss; I've looked into this rabbit hole for some time, and now it seems the problematic change I made was to add a log line in the Lua C callback (`c_prop_getter`). Now it seems there actually is an issue with logging in Lua callbacks, for me this patch fails on a segfault most of the time (but sometimes passes):

```patch
diff --git a/t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t b/t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
index e6c8f764..84922a1b 100644
--- a/t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
+++ b/t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
@@ -52,6 +52,7 @@ ok
 
 
 === TEST 2: set_property_getter() - setting host properties getter works on rewrite_by_lua
+--- ONLY
 --- wasm_modules: hostcalls
 --- http_config
     init_worker_by_lua_block {
@@ -74,9 +75,13 @@ ok
             assert(proxy_wasm.attach(_G.c_plan))
 
             local function getter(key)
+
+                print("here")
+
                 if key == "wasmx.my_property" then
                     return true, "my_value"
                 end
+
                 return false
             end
``` 

Backtrace:
```
#0  0x0000555cb8fd0f0a in log_wrapper (log=0x13e60800000067, ident=0x555cb90df8f9 "[lua] ", level=6, 
    L=0x7fcfc55f7d80) at ../ngx_lua-0.10.25/src/ngx_http_lua_log.c:104
#1  0x0000555cb8fd0ec0 in ngx_http_lua_print (L=0x7fcfc55f7d80)
    at ../ngx_lua-0.10.25/src/ngx_http_lua_log.c:86
#2  0x00007fcfc5759b93 in lj_BC_FUNCC ()
   from /code/ngx_wasm_module/work/buildroot/prefix/luajit/lib/libluajit-5.1.so.2
#3  0x0000555cb90ad7ef in ngx_proxy_wasm_properties_get (pwctx=0x555cbad04f50, path=0x7ffdc12fe650, 
    value=0x7ffdc12fe660)
    at /code/ngx_wasm_module/src/common/proxy_wasm/ngx_proxy_wasm_properties.c:1057
#4  0x0000555cb90a8a45 in ngx_proxy_wasm_hfuncs_get_property (instance=0x555cb9300900, 
    args=0x555cb9311000, rets=0x555cb9310fe0)
    at /code/ngx_wasm_module/src/common/proxy_wasm/ngx_proxy_wasm_host.c:779
#5  0x0000555cb909c39d in ngx_wavm_hfunc_trampoline (env=0x555cb93004c0, caller=0x7ffdc12fe868, 
    args=0x555cbad8f220, nargs=4, rets=0x555cbad8f280, nrets=1)
    at /code/ngx_wasm_module/src/wasm/vm/ngx_wavm_host.c:265
#6  0x00007fcfc49c3417 in wasmtime::func::c_callback_to_rust_fn::{{closure}}::hffefa239826578cc ()
   from /code/ngx_wasm_module/work/runtimes/wasmtime-14.0.3/lib/libwasmtime.so
#7  0x00007fcfc49ba82b in wasmtime::func::Func::invoke::h9cebbb299d0df8c4 ()
   from /code/ngx_wasm_module/work/runtimes/wasmtime-14.0.3/lib/libwasmtime.so
#8  0x00007fcfc49b7c64 in wasmtime::func::Caller<T>::with::h755a29576193b3ba ()
   from /code/ngx_wasm_module/work/runtimes/wasmtime-14.0.3/lib/libwasmtime.so
#9  0x00007fcfc496fba4 in wasmtime::trampoline::func::array_call_shim::h91a4e9d19e665282 ()
   from /code/ngx_wasm_module/work/runtimes/wasmtime-14.0.3/lib/libwasmtime.so
```

thibaultcha: I think overall the PR should also have tests of various Lua APIs being used within the callbacks; even if some APIs don't work, at least for circling their behavior.

lib/resty/wasmx/proxy_wasm.lua
hishamhm: > So what I encounter because of this int to ngx_int_t change is that return values (rc) from FFI calls such as C.ngx_http_wasm_ffi_set_host_property are returned as cdata types instead of number, which is quite the undesired side-effect imho, especially given int should already cover all necessary return codes, just like in the core OpenResty FFI APIs

Ah, I see. I suspect using `int` here might be working "by accident", just because we're running on little-endian systems.

If we're keeping `int` in the function definitions because it's harmless in practice, I think we need `ngx_int_t` only for the definition of the `ngx_wasm_host_prop_fn_t` function pointer typedef, because that flows the data in the opposite direction (from Lua to C), so LuaJIT compiles that into a C function stub which needs to be 64-bit aware in order to fill the return data with the correct sizeof to be consumed by C code (because otherwise I did see garbage being returned by the callback).

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
hishamhm: > For rock-solid, tight test cases, I advise using grep_error_log_out at all times, extracting the necessary patterns such that even wasmx.my_property: not found string would also come up and make this test properly fail (at least in its current test).

Yeah, I've followed some existing patterns when using `error_log` but I agree that `grep_error_log` is much better because of the checks for ordering. However, it can get pretty tricky/annoying to write a correct regexp for catching several lines in order and ensuring no unexpected repeats, etc. I'm thinking of adding a custom block handler where you would give a list of plain substrings (not regexes), and the test passes if the given substrings appear in that exact order with no additional occurrences of them. The subset of relevant lines could be filtered by a regex. It would make it lot easier to write solid tests.

EDIT: I left this comment open in the browser, went ahead, and implemented the block handler. PR incoming!

thibaultcha: Not opposed, but `grep_error_log` is still superior, because it may catch additional lines containing "the pattern" that may not be expected in the `grep_error_log_out` block... Plain substrings only solves the ordering aspect of `error_log`, but not this other issue which in my opinion, is well-worth the time spending crafting a matching regex and appropriate logs in the module & test filters.

thibaultcha: Actually I am not sure if I oppose or not. This would easily end up being used everywhere for convenience, but still not being as tight as `grep_error_log`...

hishamhm: PR ready, do check it out — it's tighter than `grep_error_log`!

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
hishamhm: No objections!

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
hishamhm: I'm able to reproduce this crash in `feat/ffi-properties-code-review` (same results as you), but not in `feat/ffi-properties-bridge-wip`. One suspicious difference is that in the bridge version, I'm calling `ngx_http_lua_set_req` explicitly to set the `r` of the `lua_State` before the callback runs. My commit log says this makes Valgrind happy, and I do recall seeing this type of error sometime during development, so I think that's how I solved this then.

hishamhm: @thibaultcha This seems to solve the issue on my end. It's a Lua equivalent of `ngx_http_lua_set_req`, which I figured out by peeking at lua-resty-core:

```
diff --git lib/resty/wasmx/proxy_wasm.lua lib/resty/wasmx/proxy_wasm.lua
index 7934c4fd..6fecff3c 100644
--- lib/resty/wasmx/proxy_wasm.lua
+++ lib/resty/wasmx/proxy_wasm.lua
@@ -12,6 +12,7 @@ local pcall = pcall
 local ffi_gc = ffi.gc
 local ffi_new = ffi.new
 local ffi_str = ffi.string
+local getfenv = getfenv
 local tostring = tostring
 local subsystem = ngx.config.subsystem
 local get_request = wasmx.get_request
@@ -89,6 +90,16 @@ do
     -- We can't create one of these on each request,
     -- so we prepare these wrappers, which in turn call a Lua function.
 
+
+    local function set_r(r)
+        local pok, exdata = pcall(require, "thread.exdata")
+        if pok and exdata then
+            exdata(r)
+        else
+            getfenv(0).__ngx_req = r
+        end
+    end
+
     local function store_c_value(r, ckey, cvalue, lvalue, is_const)
         if lvalue ~= nil then
             local value = tostring(lvalue)
@@ -107,6 +118,7 @@ do
 
     c_prop_setter = ffi.cast("ngx_wasm_host_prop_fn_t", function(r, ckey,
                                                                  cvalue)
+        set_r(r)
         local lkey = ffi_str(ckey.data, ckey.len)
         local lval
         if cvalue.data ~= nil then
@@ -131,6 +143,7 @@ do
 
     c_prop_getter = ffi.cast("ngx_wasm_host_prop_fn_t", function(r, ckey,
                                                                  cvalue)
+        set_r(r)
         local lkey = ffi_str(ckey.data, ckey.len)
 
         local pok, ok, lvalue, is_const = pcall(lua_prop_getter, lkey)
```

t/04-openresty/ffi/304-proxy_wasm_set_property_getter.t
```diff
@@ -0,0 +1,597 @@
...
+[
+    qr/\[info\] .*? \[hostcalls\] on_log/,
+    qr/\[info\] .*? wasmx.my_property: my_value/,
+]
+--- no_error_log
+[error]
+
+
+
+=== TEST 7: set_property_getter() - setting property getter at startup doesn't reset filter list
```
thibaultcha, line 303: It is not clear to me what this test is supposed to test; the startup seems to be the same as other tests, in fact this test seems functionally identical to `TEST 3 - sanity in access_by_lua`, except for `request_headers_in_access` flag which should not make any difference whatsoever? I don't see why this flag would be worth testing with host properties, but maybe am missing something. In my merge PR I marked these tests for getter/setter as `SKIP?` for now, given I'd rather get rid of them if they are redundant.

thibaultcha: @hishamhm I think this is my last item before merging; was there something I missed in this test that isn't present in any other test?

hishamhm: You're right, tests 3 and 7 in this file are currently redundant, but the reason why this one originally existed is relevant and possibly should be captured in the test suite, if not with a test of its own, at least with a comment:

I based that test off test 4 of `proxy_wasm_set_property_host.t`, which is a regression test added when fixing a bug reported by the Gateway, which was an issue when setting properties between `attach` and `start`. So I replicated that test to ensure that configuring the property getter/setter also works when done in between `attach` and `start`. However, test 3 which is checking the `access_by_lua` phase is also doing that, which rendered the test case redundant.

So in short, we should test that the various phases, and we should test that setting the getter/setter works before and after `proxy_wasm.start`. So I'd say we have two options:

* test 3 should be changed so that the getter is set _after_ `proxy_wasm.start` in the access phase, and then test 7 is renamed to "setting property getter before proxy_wasm.start works"
* test 7 is removed, we assume that if setting the getter works in later phases then it also works on access after `proxy_wasm.start`, we keep test 3 as is, and then I'd recommend adding an explicit comment in the Lua block saying something like `-- ensuring that setting the getter in between attach and start works`

thibaultcha: Great thanks! Ok so I removed the test since the `header_filter_by_lua` test case does test this (set_prop after attach/start) while all other tests do attach/set_prop/start). I think this will be merged today.

thibaultcha:
Merged.

hishamhm:
@thibaultcha Sweet! I will retarget the `.dependencies` of the Gateway PR to target `main`!

PR #441, opened 2023-11-07
[wip] feat(proxy-wasm) host-managed property getters and setters using the Lua bridge

hishamhm:


lib/resty/wasmx/proxy_wasm.lua
```diff
@@ -274,8 +279,11 @@ function _M.get_property(key)
 
     local ckey = ffi_new("ngx_str_t", { data = key, len = #key })
     local cvalue = ffi_new("ngx_str_t", { data = nil, len = 0 })
+    local rcp = ffi_new("ngx_int_t[1]", { 0 })
+
+    C.ngx_http_wasm_ffi_get_property(r, ckey, cvalue, rcp)
 
-    local rc = C.ngx_http_wasm_ffi_get_property(r, ckey, cvalue)
+    local rc = tonumber(rcp[0])
```
thibaultcha, line 286: To me the issues we've been seeing prior to this change were maybe more likely to be related to an error in stack manipulation of `L` or `cur_co` than anything else... I spent some time looking at it but it warrants more investigation since something could be wrong in the Lua bridge and we never caught it through the resolver chunk due to it having a different number of arguments/return values (`nrets` lua_resume arg).

hishamhm: > maybe more likely to be related to an error in stack manipulation of L or cur_co than anything else...

I had a similar hunch. Reading hte Lua bridge and the sources for the OpenResty functions it calls confused me a bit because I was seeing [the `nargs` value being passed](https://github.com/Kong/ngx_wasm_module/blob/main/src/common/lua/ngx_wasm_lua.c#L415) to [an argument called `nrets`](https://github.com/openresty/lua-nginx-module/blob/86bea01b244938e0ab6eda780906c06fa62c3b5c/src/ngx_http_lua_util.c#L1109).

And yeah, this commit efe1c31 was just an attempt at an ugly workaround. I was drawn to it because i was consistently seeing backtraces with `lj_BC_RET` on top:

```
(gdb) bt
#0  0x00007f9315407633 in lj_BC_RET () from /home/hisham/work/kong/ngx_wasm_module/work/buildroot/prefix/luajit/lib/libluajit-5.1.so.2
#1  0x00005573496c3e5c in ngx_http_lua_run_thread (L=0x7f9313fe0380, r=0x55734c3cf090, ctx=0x55734ade7ea0, nrets=0) at ../ngx_lua-0.10.21/src/ngx_http_lua_util.c:1185
#2  0x00005573496cd612 in ngx_http_lua_access_by_chunk (L=0x7f9313fe0380, r=0x55734c3cf090) at ../ngx_lua-0.10.21/src/ngx_http_lua_accessby.c:337
#3  0x00005573496cd1ea in ngx_http_lua_access_handler_inline (r=0x55734c3cf090) at ../ngx_lua-0.10.21/src/ngx_http_lua_accessby.c:185
#4  0x00005573496cd123 in ngx_http_lua_access_handler (r=0x55734c3cf090) at ../ngx_lua-0.10.21/src/ngx_http_lua_accessby.c:158
#5  0x00005573495bc512 in ngx_http_core_access_phase (r=0x55734c3cf090, ph=0x55734adede18) at src/http/ngx_http_core_module.c:1110
#6  0x00005573495bbb78 in ngx_http_core_run_phases (r=0x55734c3cf090) at src/http/ngx_http_core_module.c:885
#7  0x00005573495bbae1 in ngx_http_handler (r=0x55734c3cf090) at src/http/ngx_http_core_module.c:868
#8  0x00005573495cca05 in ngx_http_process_request (r=0x55734c3cf090) at src/http/ngx_http_request.c:2120
#9  0x00005573495cb076 in ngx_http_process_request_headers (rev=0x55734c202af0) at src/http/ngx_http_request.c:1498
#10 0x00005573495ca3b2 in ngx_http_process_request_line (rev=0x55734c202af0) at src/http/ngx_http_request.c:1165
#11 0x00005573495c897d in ngx_http_wait_request_handler (rev=0x55734c202af0) at src/http/ngx_http_request.c:503
#12 0x00005573495a2f15 in ngx_epoll_process_events (cycle=0x55734adc3030, timer=60000, flags=1) at src/event/modules/ngx_epoll_module.c:901
#13 0x000055734958eb39 in ngx_process_events_and_timers (cycle=0x55734adc3030) at src/event/ngx_event.c:257
#14 0x000055734959e9f8 in ngx_single_process_cycle (cycle=0x55734adc3030) at src/os/unix/ngx_process_cycle.c:323
#15 0x0000557349553c34 in main (argc=5, argv=0x7ffc47a67b38) at src/core/nginx.c:383
```

These can be seen via core dump and via Valgrind.

What's interesting is that changing the `ffi.cdef` definitions changes the values being read by "Invalid read of size 1" reported by Valgrind. If I changed the return values from `ngx_int_t` (64-bit) to `int` (32-bit), I'd get 0xffffffffffffffff back. If I use the correct 64-bit return type, I get a different, non-all-ffff number (but which is also an invalid pointer).

...so I wanted to experiment to see what happened if I avoided touching the return value of the FFI call and indeed the code seems to continue to churn along.

A quick search led me to [this old post of yours](https://www.freelists.org/post/luajit/Fixed-a-segfault-when-unsinking-64bit-pointers). I'm hoping we're just doing something silly that's putting the Lua stack in a bad state, and it's not anything like this again...

thibaultcha: Yeah I saw the same backtraces yesterday night. The `nargs` is named as such because even though the underlying argument is `nrets`, in our case "how many arguments do we want on the stack when the chunk if finished" is the equivalent of "how many arguments do we get when we do ` = ...`, which is the number of arguments we want to pass to the chunk. Hope that makes sense; it did make the bridge interface simpler.
The cause of the backtraces could be unrelated to the stack (I tried playing with the stack in openresty during my debugging yesterday), but I am not sure it could be related to that old patch... 🤔

thibaultcha: You can also build against a local OpenResty release archive with `NGX_BUILD_OPENRESTY=/home/chasum/src/openresty-releases/openresty-1.21.4.2 NGX_BUILD_CC_OPT='-DNGX_LUA_USE_ASSERT -O0' make`; so you can look deep into `ngx_http_lua_util.c` and what could be going wrong there. Ideally we either only use Lua C API or OpenResty C APIs, but not both.

hishamhm: > Ideally we either only use Lua C API or OpenResty C APIs, but not both.

Yeah, the mix in [e077a66](https://github.com/Kong/ngx_wasm_module/pull/441/commits/e077a6670e59040e084669e66e5190cd92500c73) is a bit of an exploration, to understand what's going on bit by bit.

> You can also build against a local OpenResty release archive with NGX_BUILD_OPENRESTY=/home/chasum/src/openresty-releases/openresty-1.21.4.2 NGX_BUILD_CC_OPT='-DNGX_LUA_USE_ASSERT -O0' make; so you can look deep into ngx_http_lua_util.c and what could be going wrong there.

Thanks for the pointer — I will do a bit more exploring via the Lua/C API to understand some more of what's going on and then try to switch back to the OpenResty API. That should be useful!

thibaultcha: So this new version seems to work! Locally I only had to update a test case to remove debug log greps in `305` and the suite passed (origin has conflicts with old cosmetic changes I made to `properties.c` last month, minor rebase). Good news that the prop bridge approach seems ok so far, although it seems like said bridge has a bug, or the utilization of OpenResty C APIs is incomplete perhaps?

hishamhm: (ah, further note: simply changing the `lua_resume` back to the `ngx_wasm_lua_thread_run` doesn't work, as it then makes the ngx_wasm_module tests fail.)

hishamhm: > Not looking forward to resync the OpenResty patched version, but if that's the only way to test it with the Gateway, I guess that's the way to go...

So I have backported several dozens of commits from LuaJIT's upstream `v2.1` into `kong/build/openresty/patches`... I blanket-picked all commits from April 10 2023 onwards because the LuaJIT version in OpenResty is called `LuaJIT-2.1.-20230410`. Many didn't apply cleanly, indicating quite some divergence between the OpenResty and upstream trees (and Mike's recent move to the "rolling release" model contains some pretty strong words against relying on patched versions).

Anyway, imagine my feeling when I finally got a clean build, ran the Gateway tests and they start running green... until _one_ of them fails. Then I run again, and get 4 failures as usual. Then 2 failures. Then 4 again. Meh. It was a glimmer of hope. If anything, it _feels_ like it crashes a little less on average, but it might be a biased perception from someone who wished those 99 patch files were worth something :laughsob:

Do you know if it is possible to run OpenResty nowadays with _full vanilla_ LuaJIT? It's the next thing I thought of trying, which implies diving into Bazel and the OpenResty build system (...or hosting an alternative OpenResty package with the replaced LuaJIT and just changing a single URL in Bazel... :evilgrin: (yes, I'm inventing imaginary emojis)).

hishamhm: Also, to confirm, if I add `jit.off()` at the beginning of times in `kong/init.lua`, the suite always passes. But then, it also did with the first (FFI-only) version of this commit. :\

With `jit.on()`, both the FFI and Lua-bridge versions also trigger pretty much identical failures.

thibaultcha: I do not know if it still is possible to run a vanilla LuaJIT in OpenResty...

> Also, to confirm, if I add jit.off() at the beginning of times in kong/init.lua, the suite always passes. But then, it also did with the first (FFI-only) version of this commit.

At least this seems to indicate that our stack manipulation isn't the issue, but sure, that it hints to the JIT is quite perplexing.

hishamhm:
Closed in favor of #431

PR #446, opened 2023-11-09
chore(util) replicate runtimes workflow to sdks

casimiro:
This PR proposes a few changes: 
- A new structure for accommodating Proxy-wasm SDKs compilation scripts;
  `./util/sdk.sh` and `./util/sdks/*.sh`  resemble the structure and interface of the scripts used to download and build    WebAssembly runtimes: `./util/runtime.sh` and `./util/runtimes/*.sh`.

- AssemblyScript SDK compilation script and set of tests for the example filters. 

- `./util/test.sh` only compiles SDKs when the running tests require them.


Makefile
```diff
@@ -6,6 +6,8 @@ V8 ?= 11.4.183.23
 PCRE ?= 8.45
 ZLIB ?= 1.2.13
 LUAROCKS ?= 3.9.2
+PROXY_WASM_GO_SDK ?= v0.21.0
+PROXY_WASM_ASSEMBLYSCRIPT_SDK ?= v0.0.6
```
thibaultcha, line 10: Can we do like all other variables and remove the `v` from here, insert it in the URLs where the number is used?

assets/release/Dockerfiles/Dockerfile.ubuntu-22.04
```diff
@@ -54,7 +58,9 @@ RUN apt-get install -y \
     rustup target add wasm32-unknown-unknown && \
     curl -sLO https://golang.google.cn/dl/go${GOLANG_VERSION}.linux-$(cat /TG_ARCH).tar.gz && \
     sudo rm -rf /usr/local/go && tar -C /usr/local -xzf go${GOLANG_VERSION}.linux-$(cat /TG_ARCH).tar.gz && \
-    curl -sLO https://github.com/tinygo-org/tinygo/releases/download/v$TINYGO_VERSION/tinygo_${TINYGO_VERSION}_$(cat /TG_ARCH).deb && \
+    curl -sLO https://github.com/tinygo-org/tinygo/releases/download/v${TINYGO_VERSION}/tinygo_${TINYGO_VERSION}_$(cat /TG_ARCH).deb && \
     dpkg -i tinygo_${TINYGO_VERSION}_$(cat /TG_ARCH).deb && \
+    curl -sLO https://nodejs.org/dist/v${NODEJS_VERSION}/node-v${NODEJS_VERSION}-linux-$(cat /NODEJS_ARCH).tar.xz && \
+    tar -C /usr/local --strip-components=1 -xf node-v${NODEJS_VERSION}-linux-$(cat /NODEJS_ARCH).tar.xz && \
```
thibaultcha, line 64: Could you uncomment the section in `release.yml` so that this PR can trigger a release job? That way we can test if the PR doesn't break the release pipeline.

t/03-proxy_wasm/sdks/003-assemblyscript_sdk.t
```diff
@@ -0,0 +1,111 @@
...
+--- no_error_log
+[error]
+[crit]
+
+
+
+=== TEST 2: proxy_wasm AssemblyScript SDK - auth example, authorized
+--- load_nginx_modules: ngx_http_echo_module
+--- wasm_modules: assemblyscript_auth
+--- tcp_listen: 12345
```
thibaultcha, line 37: Can we use the `SERVER_PORT2` variable here instead of hard-coding a port number? Because `gen_rand_port` will actually check for used ports iirc.

util/setup_dev.sh
```diff
@@ -179,6 +179,16 @@ if [[ -n "$NGX_WASM_RUNTIME" ]] && ! [[ -n "$NGX_WASM_RUNTIME_LIB" ]]; then
...
 
+if [[ ! -d "$DIR_PROXY_WASM_GO_SDK" ]]; then
+    notice "downloading proxy-wasm go SDK..."
+    $NGX_WASM_DIR/util/sdk.sh -S go --clean
+fi
+
+if [[ ! -d "$DIR_PROXY_WASM_ASSEMBLYSCRIPT_SDK" ]]; then
+    notice "downloading proxy-wasm AssemblyScript SDK..."
+    $NGX_WASM_DIR/util/sdk.sh -S assemblyscript --clean
+fi
```
thibaultcha, line 190: Do we have somewhere the checks we discussed at length? Namely, that if a machine does not have Node.js or TinyGo installed, these setups should be skipped? The part I was the most excited about with this PR was that we would stop requiring TinyGo (and don't add Node.js) from the list of "hard requirements"; the goal being to ease-up development onboarding for the module. Did I miss something that does this in the changeset?

util/test.sh
```diff
@@ -94,7 +94,20 @@ if [[ "$CI" == 'true' ]]; then
...
+    $NGX_WASM_DIR/util/sdk.sh -S go --build
+fi
+
+if [[ $PARAMS =~ "assemblyscript_sdk" ]]; then
+    $NGX_WASM_DIR/util/sdk.sh -S assemblyscript --build
+fi
+
+if [[ $PARAMS =~ "go_sdk" ]]; then
+    $NGX_WASM_DIR/util/sdk.sh -S go --build
+fi
```
thibaultcha, line 110: I recall we also discussed this actually needed to move _out_ of `test.sh`; only `setup_dev.sh` has the "right" to install new SDKs (if the underlying dependencies are met). The tests just rely on `skip_***` to skip a non-available SDK.

Makefile
casimiro: Good catch!

casimiro:
Thank you for the careful and detailed review @thibaultcha. I'm addressing the issues highlighted in the comments.
I'll remove the WIP label when they are addressed and the release job runs successfully.

thibaultcha (commented):
Ok looking neat now! I will give it spin.

PR #449, opened 2023-11-26
[CI] merge(*) host properties FFI getter/setter

thibaultcha:
Changes atop #431:

- Properties
    - Rename to project patterns (e.g. `_handler_pt`, `host_props`) and
      "host property getter/setter" from within the FFI. Since the feature
      can only be used through the FFI, proxy_wasm_properties know it as
      "FFI host props getter/setter".
    - Guard feature to OpenResty builds
    - Share the request pointer between getter/setter
    - Various optimizations and readability improvements (e.g.
      assertions for readers)
- FFI
    - Re-organize proxy_wasm.lua
    - Fix FFI definitions signatures (`int`)
    - `store_c_value` without `retrieve` in setter (for the reader)
    - Avoid double-logging getter/setter callback errors
    - Allow logging in callbacks (with `set_r` path from Hisham)
    - Obtain callback errors in `get/set_property` err return value,
      this error has no stacktrace hence contains some format with the
      property key.
- Hostcalls
    - Produce a trap on callback errors
- Tests
    - Trim all test cases to bare-minimum
    - Switch multiple log checks to `--- grep_error_log`
    - Test logging in callbacks
    - Test yielding behavior in callbacks
    - Test errors through FFI + filters in both getters/setters
    - Rename and reorder tests consistently between 03-proxy-wasm and
      04-openresty


hishamhm:
> Guard feature to OpenResty builds.

This is the only change I would argue against: the original design of host-defined properties kept the Lua and non-Lua sides well separated, and this was proven when I made three different implementations of the Lua side (FFI, Lua/C, Lua bridge APIs) and ngx_proxy_wasm_properties.c remained unchanged -- there was nothing FFI-specific (or even Lua-specific) about the additions there. My reasoning was that, as an Nginx module that can used with and without OpenResty, it is reasonable to think that a different non-Lua-driven Nginx application could still want to register a C function callback as a driver for dynamic host-defined properties.

I don't feel strongly about this, but I just want to point out the rationale on the design and the use-case.

thibaultcha:
If that is ever a use-case the guards can be removed (I doubt it ever will of course). In the meantime, the guards help the reader better mapping what parts of `properties.c` have to do with what other parts of the codebase, and what can be eliminated when working on vanilla Nginx builds - it's mostly reader-centric (maintainer-centric).

PR #450, opened 2023-11-27
feat(build) Add bazel rules to build ngx-wasm-rs and ngx-rust

curiositycasualty:
"This PR adds Bazel build rules to compile ngx-wasm-rs and ngx-rust that can be imported and used by Kong,
as directly calling `cargo` might break cross-compilation.

The `rules_rust` doesn't seem to parse the local redirection of depenendencies in Cargo.toml well so that
I have to manually add those dependencies. And it seems the secondary dependencies will also needs to
be in the top level Cargo.toml to make build happy. Let me know how do you feel about this workaround.

What's missing at present:
- [ ] wat feature (currently not working due to "missing CMAKE_ROOT")
- [ ] readme
- [ ] reusable workflow"

~ https://github.com/Kong/ngx_wasm_module/pull/392

This PR is based on a copy of the @fffonion fork branch that has been pushed to the @Kong repo, duplicating https://github.com/Kong/ngx_wasm_module/pull/392 while allowing for CI *in this repo* to run.


curiositycasualty:
Rebased on main.

config
```diff
@@ -102,7 +102,7 @@ if [ "$ngx_wasm_cargo_lib_name" = ngx_wasm_rs ]; then
         fi
     fi
 
-    if echo -n "$ngx_wasm_runtime_libs\n$NGX_LD_OPT" | grep -q "ngx-wasm-rs"; then
+    if echo -n "$ngx_wasm_runtime_libs\n$NGX_LD_OPT" | grep -q "ngx_wasm_rs"; then
```
curiositycasualty, line 105:
```suggestion
    if echo -n "$ngx_wasm_runtime_libs\n$NGX_LD_OPT" | grep -E -q "ngx_wasm_rs|ngx-wasm-rs"; then
```

fffonion:
That will lead to the question of if we are going to ship the two other runtimes (`wasmer` and `v8`) that need either `ngx_wasm_rs` or `ngx_rust`, other than the existing `wasmtime`, into kong or kong-ee. Which from a dicussion
from my previous PR a while ago, I assume the answer is yes. 

If the plan changes and that we aren't planning to ship the two other runtimes with kong in near future, we can hold this
progress for now. But if we do, I would suggest a go with this PR. On this repo, we will only ensure the build part, there's no need to switch the whole CI of ngx-wasm-module to build upon bazel. Additional functional tests will then need to be
added on gateway side to cover the two other runtimes.

thibaultcha:
Hmm ok. We are probably not shipping any other runtime anytime soon, but maybe we could still have this. Heads up though there will be quite some work before this can be merged based on this current state.

>  there's no need to switch the whole CI of ngx-wasm-module to build upon bazel. 

Oh well yeah that has always been totally out of the question anyway. If we add bazel it's just for fun (CONTRIBUTING.md note) or for a specific Gateway purpose (even though we probably will only ever ship with Wasmtime).

thibaultcha:
Since it seems we only need this because of how the Gateway wishes to build our Rust libs, could we move it to a repository of its own instead? We can keep the CI in this repository, and the bazel rules in a new one like `Kong/ngx_wasm_module_bazel` or something...

curiositycasualty:
> could we move it to a repository of its own instead?

If we come back around on needing this functionality, than that can be the approach we take.
Closing per [KAG-2682](https://konghq.atlassian.net/browse/KAG-2682).

PR #451, opened 2023-11-29
tests(*) switch Valgrind tests from opt-out to opt-in

thibaultcha:
Remove `--- skip_valgrind` block and helper in favor of an opt-in mode for Valgrind test cases.


hishamhm:
@thibaultcha Do you have any rules-of-thumb you've been using to decide whether a test should have Valgrind enabled? If so, perhaps it would be good idea to document them in docs/DEVELOPER.md, so we can try to apply them as well when submitting PRs.

PR #456, opened 2023-12-01
fix(proxy-wasm) use pwexec pool in ngx_proxy_wasm_maps_set

hishamhm:
Use the pwexec pool (which is per-request in lower isolation modes) rather than the persistent instance pool in ngx_proxy_wasm_maps_set.

This fixes a memory-hoarding issue observed when filters set headers on every request.


hishamhm:
This should fix the memory growth issues observed by @tonydongkong in his tests.

PR #457, opened 2023-12-01
fix(wasi) do not use instance pool on fd_write

hishamhm:
Avoid using the long-lived instance pool for a temporary buffer on fd_write. Since we allocate and deallocate deterministically within the same function (and to avoid a cross-concern dependency on proxy-wasm to get pwctx), we simply use `ngx_alloc` and `ngx_free` to produce a temporary buffer.

### DIFF
--- a/src/wasm/wasi/ngx_wasi_host.c
+++ b/src/wasm/wasi/ngx_wasi_host.c
@@ -258,7 +258,7 @@ ngx_wasi_hfuncs_fd_write(ngx_wavm_instance_t *instance,
 
     /* second pass: build string from iov buffers */
 
-    msg_data = ngx_pcalloc(instance->pool, msg_size);
+    msg_data = ngx_alloc(msg_size, instance->log);
     if (msg_data == NULL) {
         rets[0] = (wasm_val_t) WASM_I32_VAL(WASI_ERRNO_EFAULT);
         return NGX_WAVM_OK;
@@ -290,6 +290,8 @@ ngx_wasi_hfuncs_fd_write(ngx_wavm_instance_t *instance,
         ngx_log_error(level, instance->log_ctx.orig_log, 0, "%*s",
                       msg_size, msg_data);
     }
+
+    ngx_free(msg_data);
 
     rets[0] = (wasm_val_t) WASM_I32_VAL(WASI_ERRNO_SUCCESS);
     return NGX_WAVM_OK;


hishamhm:
@thibaultcha quick Q about the codebase:

While looking for a different pool accessible from the `instance` struct to use in this PR, I've seen `instance->data` (from a `ngx_wavm_instance_t *instance`) being cast to `ngx_proxy_wasm_instance_t *` in `src/common/proxy_wasm/ngx_proxy_wasm.h` and to `ngx_http_wasm_req_ctx_t *rctx` in various places at `src/http/ngx_http_wasm_host.c` and `src/wasm/ngx_wasm_core_host.c`. From what I could tell these are different structures, so I assume `instance->data` is being set to different things at different times, is that assumption correct?

Without following through the whole execution logic, I wasn't able to tell what's the state of the `instance->data` pointer when the `ngx_wasi_host.c` functions run, so I just went with the simple ngx_alloc/ngx_free approach in this PR (since that is also used in a few other places of the codebase). Do we have a shorter-lived pool accessible by the WASI funcs?

(All code in `src/wasm/wasi/ngx_wasi_host.c` is currently HTTP-agnostic, so casting `instance->data` any way would require some intrusive #include-ing...)

thibaultcha:
Right, I'll probably elaborate a bit more than I think I need to but maybe it'll be helpful somehow.

So from the point of view of the codebase what we see when looking for `instance->data` in a hostcall depends on the current user of that instance; or in other words: there are two different users of wasm instances currently in the module: `wasm_call` and `proxy_wasm`.
The structure that encompass both `wasm_call` and `proxy_wasm` is `ngx_wasm_ops`, the "operations scheduler". When running, this "operations scheduler" invokes the `ngx_wavm` API for creating an instance with a host data pointer, `instance->data`.
The host data pointer for `wasm_call` is defined [here at creation time](https://github.com/Kong/ngx_wasm_module/blob/main/src/wasm/ngx_wasm_ops.c#L348), hence inherited to be `rctx`. For `proxy_wasm`, we devised the isolation mechanism so instances can be reused, hence the host data pointer is always forcibly set in ngx_proxy_wasm.c via `ngx_wavm_instance_set_data()`, [here](https://github.com/Kong/ngx_wasm_module/blob/main/src/common/proxy_wasm/ngx_proxy_wasm.c#L1680).
So, two different "codebase users" of `ngx_wavm` have two different stacktraces leading outside of their hostcalls, thus can assume the type of their host data pointer:
- The `ngx_proxy_wasm_instance_t *` cast in the `proxy_wasm` codebase user
- And `ngx_http_wasm_req_ctx_t *rctx` in `ngx_http_wasm_host.c`, the "http hostcalls" for `wasm_call`. (The "http hostcalls" being totally unused at the moment since we only focus on Proxy-Wasm SDK support and not our own "WasmX" SDK.

> Without following through the whole execution logic, I wasn't able to tell what's the state of the instance->data pointer when the ngx_wasi_host.c functions run

Yes, the above assumptions fail when considering a hostcalls interface shared between multiple `ngx_wavm` codebase users, aka the WASI hostcalls...

> (All code in src/wasm/wasi/ngx_wasi_host.c is currently HTTP-agnostic, so casting instance->data any way would require some intrusive #include-ing...)

I think the long-term ideal is to make use of `ngx_wasm_subsys` in the host call pointers; it represents an encapsulated "subsystem context", stream or http. It already exists in `rctx->env`, but was only added recently so most of the codebase isn't using it.
Then for Proxy-Wasm hostcalls (storing `ngx_proxy_wasm_instance_t *` at the moment), since the `proxy_wasm` caller already sets `rctx->data = pwctx`, I think we could maybe get away with storing `rctx` in `instance->data` instead of `ictx`... If we could do that, then we could set `instance->data` to `&rctx->env`, and fully encapsulate the host data pointer for "HTTP Wasm", "Proxy-Wasm", and "WASI" hostcalls.
All hostcall interfaces could interpret their host pointer with certainty; the WASI hostcalls could probably rely on `rctx->env` alone, given it has the connection pool, log, etc, so there's probably no need `ngx_http` code at all in it if we do this.

hishamhm:
@thibaultcha Thanks for the writeup! I followed it along with the struct definitions and it all made sense. Having `instance->data` point to a `ngx_wasm_subsys_t` sounds like a good solution, as it will make `pwctx` accessible to the proxy-wasm context, `rctx` available to the wasm_call context and things like `connection->pool` and `connection->log` generically available to the WASI context.

For the current PR in the short term, do we just go with ngx_alloc/ngx_free as implemented here?

thibaultcha:
Yes, we can go with this patch for now and I'll add a TODO on it merging it.

thibaultcha:
Done.

PR #458, opened 2023-12-04
chore(sdk) separate build and install of .wasm examples

hishamhm:
The SDK installer scripts are not installing the example filters into their target locations when it detects that the sources haven't changed via the hash file. This patch splits the build and install steps to fix this behavior.

The checks are still not as ideal (we're still not checking the sources and targets for each example individually, so if we get a compilation failure halfway through the build we're still going to end up in an inconsistent state), but it is a step in the right direction:

* check that some .wasm files are actually present when deciding to skip the build
* always install .wasm files into the target location, even when build is skipped

### DIFF
--- a/util/sdks/go.sh
+++ b/util/sdks/go.sh
@@ -66,10 +66,11 @@ build_go_sdk() {
     if [[ -d "$DIR_PATCHED_PROXY_WASM_GO_SDK" \
           && -f "$DIR_PATCHED_PROXY_WASM_GO_SDK/.hash" \
           && $(cat "$DIR_PATCHED_PROXY_WASM_GO_SDK/.hash") == $(echo $hash_src)
-          && -z "$clean" ]];
+          && -z "$clean" ]] && \
+       find $DIR_PATCHED_PROXY_WASM_GO_SDK -name '*.wasm' | grep -q .
     then
         notice "Go examples already built"
-        exit
+        return
     fi
 
     rm -rf $DIR_PATCHED_PROXY_WASM_GO_SDK
@@ -105,7 +106,11 @@ EOF
         notice "compiling Go examples..."
 
         make build.examples || exit 0
+    popd
+}
 
+install_go_sdk_examples() {
+    pushd $DIR_PATCHED_PROXY_WASM_GO_SDK
         cd examples
 
         find . -name "*.wasm" | while read f; do
@@ -128,4 +133,5 @@ if [ "$mode" = "download" ]; then
 
 else
     build_go_sdk "$version" "$clean"
+    install_go_sdk_examples
 fi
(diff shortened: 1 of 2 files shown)


util/sdks/go.sh
```diff
@@ -66,10 +66,11 @@ build_go_sdk() {
     if [[ -d "$DIR_PATCHED_PROXY_WASM_GO_SDK" \
           && -f "$DIR_PATCHED_PROXY_WASM_GO_SDK/.hash" \
           && $(cat "$DIR_PATCHED_PROXY_WASM_GO_SDK/.hash") == $(echo $hash_src)
-          && -z "$clean" ]];
+          && -z "$clean" ]] && \
+       find $DIR_PATCHED_PROXY_WASM_GO_SDK -name '*.wasm' | grep -q .
     then
         notice "Go examples already built"
-        exit
+        return
```
thibaultcha, line 73: Can this stay `exit` like `assemblyscript.sh`, or the later be switch to also `return`? Aiming at keeping both scripts identical.

hishamhm: Both should be `return`. It was a mistake on my part.

util/sdks/go.sh
```diff
@@ -128,4 +133,5 @@ if [ "$mode" = "download" ]; then
 
 else
     build_go_sdk "$version" "$clean"
+    install_go_sdk_examples
```
thibaultcha, line 136: Shouldn't `install_go_sdk_examples` also be invoked in the "download" mode?

thibaultcha: Ditto: download mode

hishamhm: No, because AFAICT that's for downloading the SDK only. The idea was only to factor out the install step out of build so that one could skip rebuild without skipping install (hence return instead of exit)

thibaultcha: Hmm right ok I looked at the full script and I see now, it isn't exactly like `runtime.sh`. Ok then! I'll merge with the `return` change.

thibaultcha: Also just to leave a note on it: I'm not sure we need this "download" mode for SDKs... It's harmless for now but the more SDKs we add, the more likely we'll prune it out because it'll keep the scripts lighter imho.

util/sdks/go.sh
thibaultcha: Ok I see now, they really need to be `return` indeed. Another question for you:

```sh
make build.examples || exit 0
```

Does this look right? Should we silently fail if `make build.examples` errors-out? 🤔

thibaultcha:
Merged it with the `return` fix

util/sdks/go.sh
hishamhm: > Does this look right to you? Should we silently fail if make build.examples errors-out?

I was tempted to change it to `exit 1` and even did it for a bit, but backtracked last minute because I didn't really chase the history to see how that ended up there, and given the explicit `0` there might have been a reason I don't know/remember, so in the end I wanted to keep only the necessary fix in this PR and not go after other unrelated improvements.

PR #462, opened 2023-12-07
fix(*) fixes for ngx_wavm and ngx_proxy_wasm

thibaultcha:
All fixes were uncovered one-by-one after each previous fix.


casimiro:
Nice one.

PR #463, opened 2023-12-08
chore(ci) fix large CI workflow file

casimiro:
[This](https://github.com/Kong/ngx_wasm_module/actions/runs/7160665577) successful run demonstrates the fix.

### DIFF
--- a/.github/workflows/ci-large.yml
+++ b/.github/workflows/ci-large.yml
@@ -139,7 +139,7 @@ jobs:
         v8: [11.4.183.23]
         include:
           - label: old_nginx
-            os: [ubuntu-latest]
+            os: ubuntu-latest
             cc: clang-15
             ngx: 1.21.6
             runtime: wasmtime


thibaultcha:
Glad that you found the root cause, awesome. How could we fix it without removing the ability to run `job-build-tests.yml` with different `os`? E.g. `macos-latest` in the near future?

casimiro:
While addressing the issue I concluded that the problem was the single-entry bracket-list and that having more than one entry, e.g. `os: [ubuntu-latest, macos-latest]`, wouldn't trigger the error.

But I actually haven't tested adding more entries to that list to confirm. I'll do that.

If having more items on that list also solves the issue, we can simply put the variable `os` back to the matrix containing `ubuntu-latest` and the added OS, e.g. `os: [ubuntu-latest, macos-latest]`, when we want to add a new OS.

.github/workflows/ci-large.yml
```diff
@@ -139,7 +139,7 @@ jobs:
         v8: [11.4.183.23]
         include:
           - label: old_nginx
-            os: [ubuntu-latest]
+            os: [ubuntu-latest, macos-latest]
```
thibaultcha, line 142: Hmm iirc we can't have array values in `include`?

casimiro: Hm, I think you are right. I'll update the commit accordingly.

casimiro:
The fix can be verified in [this](https://github.com/Kong/ngx_wasm_module/actions/runs/7171060391) successful run.

PR #468, opened 2023-12-13
fix(wasmtime/macos) disable Mach integration falling back to posix signals

casimiro:


t/01-wasm/002-shm_queue_sanity.t
```diff
@@ -14,6 +14,7 @@ __DATA__
 
 === TEST 1: queue - with minimum size
 --- valgrind
+--- skip_eval: 5: ($::osname =~ m/darwin/ && $::archname =~ /arm64/)
```
thibaultcha, line 17: I don't know if you follow the changes in the main branch but since recently we don't specify the number of skipped tests in this block anymore.

t/01-wasm/002-shm_queue_sanity.t
```diff
@@ -25,3 +26,20 @@ __DATA__
 --- no_error_log
 [error]
 [crit]
+
+
+
+=== TEST 2: queue - with minimum size, macOS (arm64)
```
thibaultcha, line 32: I think that instead of duplicating the test, we should have the size value be conditional based on the OS/arch with `main_config eval` blocks. We do this several times already in the tests (and in recent changes in main as well).

t/01-wasm/002-shm_queue_sanity.t
casimiro: Yes, I follow the main branch I noticed those changes.

However, since I've seen many instances of `skip_eval` still specifying the number of tests to be skipped in the [commit](https://github.com/Kong/ngx_wasm_module/commit/a278bb7ce5da7f26f6faeb98352890313af05e9b#diff-80df4c463a1ccc8ebbfd5b1f3ec4cb380e77db0fe55dab23503ac3820cb81740L418) introducing the improvement, I assumed it didn't apply to `skip_eval`.

After reading you comment I tried removing the number of skipped tests from a few `skip_eval` instances but they didn't work.

thibaultcha: Hmmm ok then, let's just keep them like this for now and we'll look later what goes wrong without the number.

t/01-wasm/directives/003-shm_directives.t
```diff
@@ -60,7 +60,7 @@ qr/\[emerg\] .*? \[wasm\] invalid shm size "1x"/
         shm_kv my_kv_1 8192;
     }
 --- error_log eval
-qr/\[emerg\] .*? \[wasm\] shm size of 8192 bytes is too small, minimum required is 12288 bytes/
+qr/\[emerg\] .*? \[wasm\] shm size of 8192 bytes is too small, minimum required is (12288|49152) bytes/
```
thibaultcha, line 63: I think we may be better off with just `\d+` here and in the likes from now on... Perhaps we shouldn't hard-code minimum page sizes of hosts which we have no way of knowing for all hosts when making updates to the test suite. For example if we have a test case that hard-codes size "XYZ", and we then make updates to the code that changes what this "XYZ" output is. We have no way of knowing what the new "XYZ" output is for all hosts, unless we try the test on each host with various page sizes... For instance, my local host machine is a Linux desktop with page size `4096`, but how would I know the page size of all the other hosts this test would run onto? Surely GHA is not going to tell me unless I spend extra time extracting the specific logs from this specific test. Hopefully that makes sense, I think we should be more lose.

casimiro: Yes, it does make sense. It's indeed laborious to maintain these tests preserving the actual numbers.

At first I was committed to preserving them as they help documenting the non trivial underlying behaviour asserted in some test cases, e.g. the ones involving eviction;  this [one](https://github.com/Kong/ngx_wasm_module/blob/main/t/03-proxy_wasm/hfuncs/shm/004-kv_eviction.t#L84) being a good example.

But the documentation of the related directives can also be expanded to include the knowledge being removed from the test cases, if we agree on the need to do so.

t/01-wasm/002-shm_queue_sanity.t
```diff
@@ -3,6 +3,10 @@
 use strict;
 use lib '.';
 use t::TestWasm;
+use POSIX qw(sysconf _SC_PAGESIZE);
+
+our $page_size = sysconf(_SC_PAGESIZE);
```
casimiro, line 8: `TestWasm.pm` seems like a good place for `$page_size` declaration and definition, although only `shm` tests will potentially benefit from it.

thibaultcha:
I am noticing that the tests are taking 4hrs to run this macos job... Probably what you were saying with the compiler stuff, but in any case it is unacceptably long, it should be under 30/45min *at most*. Perhaps we should restrict the number of tests run just like we do with Valgrind.

.github/workflows/ci-large.yml
```diff
@@ -44,6 +44,34 @@ jobs:
             hup: no_hup
             ssl: ssl
             module_type: static
+          # macOS - Wasmer (full)
+          - runtime: wasmer
+            wasmtime: 3.1.1
```
thibaultcha, line 49:
```suggestion
            wasmer: x.y.z
```

.github/workflows/ci-large.yml
```diff
@@ -44,6 +44,34 @@ jobs:
...
+          - runtime: wasmtime
+            wasmtime: 14.0.3
+            os: macos-latest
+            cc: clang
+            ngx: 1.25.3
+            ssl: ssl
+            debug: debug
+            hup: hup
+            path: t/01-wasm
+          # macOS - V8 (full)
```
thibaultcha, line 66: Should we test the whole suite with V8 or is 01-wasm enough?

.github/workflows/ci.yml
```diff
@@ -43,6 +43,7 @@ jobs:
         debug: [debug, no_debug]
         hup: [no_hup]
         module_type: ["static"]
+        path: [""]
```
thibaultcha, line 46: Can it be omitted if `path` is not required?

auto/cargo
```diff
@@ -57,10 +57,22 @@ if [ $ngx_wasm_rc = 0 ]; then
...
+                ;;
+
+                *)
+                    ngx_feature_libs="\
+                        -Wl,-rpath=$ngx_wasm_cargo_target_dir \
+                        -Wl,-rpath=$NGX_PREFIX/lib \
+                        $ngx_feature_libs"
+                ;;
+
+            esac
```
thibaultcha, line 75: What is the purpose of this switch? I can't spot a difference between the 2 branches.

t/01-wasm/002-shm_queue_sanity.t
thibaultcha: Yes, moving it to `TestWasm.pm` makes sense to me, even `min_shm_size` if it makes sense to you.

t/TestWasm.pm
```diff
@@ -263,6 +265,12 @@ add_block_preprocessor(sub {
                 $timeout += 30;
                 $timeout = max($timeout, 30);
             }
+
+        } else {
+            if ($osname =~ m/darwin/ && $archname =~ m/arm64/) {
```
thibaultcha, line 270: This could be an "else if" instead of else, if?

util/runtime.sh
```diff
@@ -145,6 +145,21 @@ if check_target "$TARGET_DIR"; then
...
+            # be disabled in favor of posix signals.
+            #
+            # Patching is required until a new version of Wasmtime is released
+            # containing the commit `1c7a07b`.
+
+            notice "installing wasmtime on macOS, forcing 'build' mode"
+            MODE="build"
+        fi
+    ;;
+esac
```
thibaultcha, line 161: If we bump Wasmtime, this could be removed, right? What were the issues with bumping Wasmtime and should we merge a bump before merging this PR?

casimiro: Yeah, we can bump Wasmtime version before merging this PR. Bumping Wasmtime version implies updating a few backtrace patters in `t/01-wasm/directives/006-backtraces_directive.t` and some tests in `t/01-wasm/directives/008-flag_directive_wasmtime.t` to conform with new configuration rules adopted by Wasmtime in recent versions.

.github/workflows/ci-large.yml
casimiro: oh, thx! good catch.

.github/workflows/ci-large.yml
casimiro: Yeah, 01-wasm should be enough.

.github/workflows/ci.yml
casimiro: Good catch.

auto/cargo
casimiro: Clang expects `-rpath` to be provided as `-rpath,/target_dir` while gcc expects it to be provided as `-rpath=/target_dir`.

t/TestWasm.pm
casimiro: Indeed.

casimiro:
The WIP label is due to the dependency on #505.

t/03-proxy_wasm/hfuncs/132-proxy_dispatch_http_ssl.t
```diff
@@ -1,3 +1,4 @@
+
```
thibaultcha, line 1:
```suggestion
```

t/03-proxy_wasm/hfuncs/131-proxy_dispatch_http_timeouts.t
```diff
@@ -97,6 +98,7 @@ qr/(\[error\]|Uncaught RuntimeError|\s+).*?dispatch failed: tcp socket - timed o
 
 
 === TEST 4: proxy_wasm - dispatch_http_call() write timeout
+--- skip_eval: 4: $::osname =~ m/darwin/
```
thibaultcha, line 101: Should we also add the reason for skipping these two tests on macos?

t/TestWasmX.pm
```diff
@@ -270,6 +277,10 @@ add_block_preprocessor(sub {
                 $timeout += 30;
                 $timeout = max($timeout, 30);
             }
+
+        } elsif ($osname =~ m/darwin/ && $archname =~ m/arm64/) {
+            $timeout += 30;
+            $timeout = max($timeout, 30);
```
thibaultcha, line 283: With many intermittent tests removed from the macos suite, is this increase still necessary for sure?

.github/workflows/ci.yml
```diff
@@ -132,6 +132,35 @@ jobs:
...
+          # macOS - V8
+          - runtime: v8
+            v8: 12.0.267.17
+            os: macos-latest
+            cc: clang
+            ngx: 1.25.4
+            ssl: ssl
+            debug: debug
+            hup: no_hup
+            path: t/01-wasm
```
thibaultcha, line 163: I think this also needs the `carryforward` update for coverage.

auto/cargo
```diff
@@ -57,10 +57,22 @@ if [ $ngx_wasm_rc = 0 ]; then
...
+                        $ngx_feature_libs"
+                ;;
+
+                *)
+                    ngx_feature_libs="\
+                        -Wl,-rpath=$ngx_wasm_cargo_target_dir \
+                        -Wl,-rpath=$NGX_PREFIX/lib \
+                        $ngx_feature_libs"
+                ;;
+
```
thibaultcha, line 74:
```suggestion
```

auto/cargo
```diff
@@ -57,10 +57,22 @@ if [ $ngx_wasm_rc = 0 ]; then
...
-                -Wl,-rpath=$NGX_PREFIX/lib \
-                $ngx_feature_libs"
+            case "$NGX_PLATFORM" in
+                Darwin:*)
+                    ngx_feature_libs="\
+                        -Wl,-rpath,$ngx_wasm_cargo_target_dir \
+                        -Wl,-rpath,$NGX_PREFIX/lib \
+                        $ngx_feature_libs"
+                ;;
+
```
thibaultcha, line 67:
```suggestion
```

auto/cargo
```diff
@@ -75,7 +87,17 @@ if [ $ngx_wasm_rc = 0 ]; then
     if [ $ngx_found = yes ]; then
 
         if [ "$ngx_wasm_cargo_crate_type" = cdylib ]; then
-            ngx_wasm_cargo_cdylib="$ngx_wasm_cargo_target_dir/lib$ngx_wasm_cargo_lib_name.so"
+            case "$NGX_PLATFORM" in
+                Darwin:*)
+                    ngx_wasm_cargo_cdylib="$ngx_wasm_cargo_target_dir/lib$ngx_wasm_cargo_lib_name.dylib"
+                ;;
+
```
thibaultcha, line 94:
```suggestion
```

auto/cargo
```diff
@@ -75,7 +87,17 @@ if [ $ngx_wasm_rc = 0 ]; then
...
+                *)
+                    ngx_wasm_cargo_cdylib="$ngx_wasm_cargo_target_dir/lib$ngx_wasm_cargo_lib_name.so"
+                ;;
+
```
thibaultcha, line 98:
```suggestion
```

.github/workflows/ci.yml
casimiro: Good catch, thx!

t/03-proxy_wasm/hfuncs/131-proxy_dispatch_http_timeouts.t
casimiro: Good call.

t/03-proxy_wasm/hfuncs/132-proxy_dispatch_http_ssl.t
casimiro: Embarrassing. Thx for spotting.

t/TestWasmX.pm
casimiro: Indeed, it's not needed anymore.

thibaultcha:
It seems colima is not working on macOS 14 runners anymore. It's not installed by default anymore (see CI run). I tried installing it with `brew install [docker, ...] colima` on the previous line, but it didn't work either.

thibaultcha:
Fixed the colima thing by using macos-13 and installing the colima brew.

PR #470, opened 2023-12-20
feat(http) support sending local response in header_filter

hishamhm:
Marked it as `feat` rather than `fix` because it can be seen as expanding our proxy-wasm support coverage.

Keeping this at Draft stage because, while the changes here do resolve the issues I was experiencing in the filter I was testing, the test failure in `t/02-http/hfuncs/004-local_response.t` is real: looks like setting `r->err_status` as I'm doing here does preserve the status code of the local response produced during `on_http_response_headers`, but it ends up also overriding the custom "reason" string, so that's no good. Any thoughts on alternative solutions for this are welcome!

### DIFF
--- a/src/http/ngx_http_wasm_filter_module.c
+++ b/src/http/ngx_http_wasm_filter_module.c
@@ -67,6 +67,7 @@ static ngx_int_t
 ngx_http_wasm_header_filter_handler(ngx_http_request_t *r)
 {
     ngx_int_t                  rc = NGX_ERROR;
+    ngx_int_t                  frc = NGX_OK;
     ngx_http_wasm_req_ctx_t   *rctx = NULL;
 
     dd("enter");
@@ -87,6 +88,10 @@ ngx_http_wasm_header_filter_handler(ngx_http_request_t *r)
 
     rctx->entered_header_filter = 1;
 
+    if (rctx->local_resp_status) {
+        ngx_http_wasm_discard_local_response(rctx);
+    }
+
     if (ngx_http_wasm_produce_resp_headers(rctx) != NGX_OK) {
         goto done;
     }
@@ -95,6 +100,11 @@ ngx_http_wasm_header_filter_handler(ngx_http_request_t *r)
                              NGX_HTTP_WASM_HEADER_FILTER_PHASE);
 
     dd("ops resume rc: %ld", rc);
+
+    frc = ngx_http_wasm_flush_local_response(rctx);
+    if (frc != NGX_DECLINED) {
+        rc = frc;
+    }
 
     if (r->err_status) {
         /* previous unhandled error before resuming header_filter */
@@ -160,6 +170,8 @@ ngx_http_wasm_body_filter_handler(ngx_http_request_t *r, ngx_chain_t *in)
         }
     }
 
+    rctx->entered_body_filter = 1;
+
     ngx_http_wasm_body_filter_resume(rctx, in);
 
     if (rctx->resp_buffering) {
@@ -170,7 +182,10 @@ ngx_http_wasm_body_filter_handler(ngx_http_request_t *r, ngx_chain_t *in)
             goto done;
         case NGX_OK:
             /* chunk in buffers */
-            ngx_wasm_assert(rctx->resp_bufs);
+            if (!rctx->resp_bufs) {
+                /* 0-byte body was buffered; terminate */
+                return NGX_ERROR;
+            }
 
             if (!rctx->resp_chunk_eof) {
                 /* more to come; go again */
--- a/t/lib/proxy-wasm-tests/hostcalls/src/filter.rs
+++ b/t/lib/proxy-wasm-tests/hostcalls/src/filter.rs
@@ -139,6 +139,19 @@ impl HttpContext for TestHttp {
 
     fn on_http_response_body(&mut self, len: usize, eof: bool) -> Action {
         info!("[hostcalls] on_response_body, {} bytes, eof: {}", len, eof);
+
+        if let Some(max) = self.body_max {
+            self.body_seen += len;
+            if self.body_seen > max {
+                self.set_http_response_body(max, 0, "".as_bytes());
+                if max == 0 {
+                    return Action::Pause;
+                } else {
+                    return Action::Continue;
+                }
+            }
+        }
+
         self.exec_tests(TestPhase::ResponseBody)
     }
 
--- a/src/http/ngx_http_wasm.h
+++ b/src/http/ngx_http_wasm.h
@@ -72,6 +72,8 @@ struct ngx_http_wasm_req_ctx_s {
     unsigned                           entered_content_phase:1; /* entered content handler */
     unsigned                           exited_content_phase:1;  /* executed content handler at least once */
     unsigned                           entered_header_filter:1; /* entered header_filter handler */
+    unsigned                           entered_body_filter:1;   /* entered body_filter handler */
+    unsigned                           flushed_local_resp:1;    /* local response can be flushed only once */
 
     unsigned                           in_wev:1;                /* in wev_handler */
     unsigned                           resp_content_chosen:1;   /* content handler has an output to produce */
--- a/src/http/ngx_http_wasm_local_response.c
+++ b/src/http/ngx_http_wasm_local_response.c
@@ -47,7 +47,7 @@ ngx_http_wasm_stash_local_response(ngx_http_wasm_req_ctx_t *rctx,
 
     dd("enter");
 
-    if (rctx->entered_header_filter) {
+    if (rctx->entered_body_filter) {
         return NGX_ABORT;
     }
 
@@ -188,6 +188,10 @@ ngx_http_wasm_flush_local_response(ngx_http_wasm_req_ctx_t *rctx)
         return NGX_DECLINED;
     }
 
+    if (rctx->flushed_local_resp) {
+        return NGX_ERROR;
+    }
+
     ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "wasm flushing local_response");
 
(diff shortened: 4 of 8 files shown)


hishamhm:
Also, I have another concern: 

The test cases I changed/added here were already green before the filter was actually working (it only started working with the `r->err_status` change), so I tried to come up with a test case to specifically represent the filter behavior: a 406 upstream response that was changed into a body-less 403 response by the filter using `send_local_response`.

This attempt however, not only doesn't work on this branch, but it weirdly hangs until a timeout:

```diff
diff --git t/03-proxy_wasm/hfuncs/102-proxy_send_local_response.t t/03-proxy_wasm/hfuncs/102-proxy_send_local_response.t
index b849ad6d..916242d3 100644
--- t/03-proxy_wasm/hfuncs/102-proxy_send_local_response.t
+++ t/03-proxy_wasm/hfuncs/102-proxy_send_local_response.t
@@ -323,6 +323,42 @@ should invoke on_log
 
 
 
+=== TEST 13: proxy_wasm - send_local_response() can override an error status
+--- load_nginx_modules: ngx_http_echo_module
+--- wasm_modules: hostcalls
+--- http_config eval
+qq{
+    upstream test_upstream {
+        server unix:$ENV{TEST_NGINX_UNIX_SOCKET};
+    }
+
+    server {
+        listen unix:$ENV{TEST_NGINX_UNIX_SOCKET};
+
+        location /t {
+            return 406 '[original response body]';
+        }
+    }
+}
+--- config
+    location /t {
+        proxy_wasm hostcalls 'on=response_headers \
+                              test=/t/send_local_response/status/403';
+        proxy_pass http://test_upstream/t;
+    }
+--- error_code: 403
+--- response_body
+--- error_log eval
+[
+    qr/testing in "ResponseHeaders"/,
+    qr/\[info\] .*? on_log/
+]
+--- no_error_log
+[error]
+[crit]
+
+
+
 === TEST 14: proxy_wasm - send_local_response() from on_http_response_headers with content
 should produce a response
 should invoke on_log
diff --git t/lib/proxy-wasm-tests/hostcalls/src/types/test_http.rs t/lib/proxy-wasm-tests/hostcalls/src/types/test_http.rs
index eb8476f6..aea3314d 100644
--- t/lib/proxy-wasm-tests/hostcalls/src/types/test_http.rs
+++ t/lib/proxy-wasm-tests/hostcalls/src/types/test_http.rs
@@ -76,6 +76,10 @@ impl TestHttp {
             /* send_local_response */
             "/t/send_local_response/status/204" => test_send_status(self, 204),
             "/t/send_local_response/status/300" => test_send_status(self, 300),
+            "/t/send_local_response/status/403" => {
+                test_send_status(self, 403);
+                return Action::Pause; /* this was an attempt to fix the behavior but made no difference */
+            },
             "/t/send_local_response/status/1000" => test_send_status(self, 1000),
             "/t/send_local_response/headers" => test_send_headers(self),
             "/t/send_local_response/body" => test_send_body(self),
```

When running the test above, the status code gets replaced and the `Content-Length` is being changed to `0`, but one of the errors I get is that the upstream body is still being produced in the response.

In the coraza filter, they [overwrite the body with null characters](https://github.com/corazawaf/coraza-proxy-wasm/blob/419790697c8fb1f5728642f1314217ab72a716cf/wasmplugin/plugin.go#L526) if OnHttpResponseBody is executed after a SendLocalResponse was issued to interrupt the response, but the comment linked here seems to indicate they didn't expected to have to do so (since this comment is in their main branch, I guess this was based out of Envoy's behavior...).

This was surprising to me as well, since I was perhaps naively expecting that SendLocalResponse would behave like Kong PDK's `kong.response.exit`, which replaces the whole body and finishes further processing with an `ngx.exit`.

It's not clear to me if the correct semantics of SendLocalResponse with a null body argument is supposed to mean "send a response without a body" or "send a response keeping the upstream body", so I wonder if the test as written in this comment's diff _is supposed_ to pass or not. 

In any case, I guess it isn't supposed to be hanging, so there's got to be something else still amiss in this branch.

thibaultcha:
> setting r->err_status as I'm doing here does preserve the status code of the local response produced during on_http_response_headers, but it ends up also overriding the custom "reason" string

I might be missing something but, why do we have to preserve the status code of the first-produced response? Can't the second response have a different status code? I am not sure why we want this behavior or why we would want to set `r->err_status` at all 🤔

thibaultcha:
> In any case, I guess it isn't supposed to be hanging, so there's got to be something else still amiss in this branch.

I think what goes on in this case is the `Content-Length` is properly set to `0`, but the body upstream buffers are not dismissed and Nginx still tries to buffer them down to the client (who isn't expecting none anytime soon since the `CL` says no body). I think this because running the test with debug logs I can see that the `on_response_body` step is seeing some bytes... That is my first guess, although how to properly dismiss the response buffers I would have to check the filter logic again, it's always tricky.

thibaultcha:
So I think that I wasted some time trying to make your branch work @hishamhm, but that in hindsight my initial intuition was correct: modifying the response once in `header_filter` is going to be trickier than the branch expects.

The module makes a point to carefully select the response content once and only once, then clearly produce it (hence the whole stashed response thing, for example). Once `ngx_http_output_filter` has been invoked, it should not be invoked again. In other words, the local response can only be flushed **once**.

I'd like to go back to the Kong PDK `response.exit` behavior you talked about, and found that it is itself a higher-level abstraction than our current implementation of local responses in ngx_wasm_module: 

- For the response body: It sets a `ngx.ctx.response_body` variable ([here](https://github.com/Kong/kong/blob/master/kong/pdk/response.lua#L778)) that is then used in `body_filter` to override the current chunk ([there](https://github.com/Kong/kong/blob/master/kong/init.lua#L1664)).
- For the response status, it seems to set `ngx.status` [early on](https://github.com/Kong/kong/blob/master/kong/pdk/response.lua#L658), but then it exits with `ngx.exit(ngx.OK)` [here](https://github.com/Kong/kong/blob/master/kong/pdk/response.lua#L805), so does this response status code update actually work from `header_filter`?

To sum it up, updating the response code and body from within a response handler like `header_filter`, we need to stash the "last minute updated code/body" in some C-side variables that can be interpreted in the same handler, last minute. This is something Kong does because it sits at a higher-level abstraction than ngx_wasm_module utilities (it is the application, not the framework). If we were to implement the same behavior in the module, we should also implement it in the application, and not the framework/utils library (i.e. in the `proxy_wasm.c` code, imho).

But with all the said, I still have doubts with the response status code because I tried making the test in your last comment pass (HTTP `406` to `403`) by updating `r->headers_out.status = 403` in `header_filter`, and it did not succeed. Hence my above comment on the `kong.response.exit` status code update from `header_filter`: does it actually work? I should dig a potential test case but wanted to drop this update first.

src/http/ngx_http_wasm_filter_module.c
```diff
@@ -96,6 +101,11 @@ ngx_http_wasm_header_filter_handler(ngx_http_request_t *r)
 
     dd("ops resume rc: %ld", rc);
 
+    frc = ngx_http_wasm_flush_local_response(rctx);
```
thibaultcha, line 104: This invocation needs to go away; we should not give the false impression that calling `flush_local_response` twice is possible; in fact I have added an item on my TODO list to catch this from within the function some time soon. If the intention is to update the status code then we should update the status code, and not rely on this function to do it for us.

thibaultcha: The status code update could be a branch in the `send_local_response` host function of the Proxy-Wasm interface instead.

thibaultcha:
I don't know the answer to any of the open questions, sorry.

thibaultcha:
For now I propose to trade you [my attempt at this PR] for [[this V8 build issue in the CentOS 7 image](https://github.com/Kong/ngx_wasm_module/pull/481)].

t/lib/proxy-wasm-tests/hostcalls/src/filter.rs
```diff
@@ -139,6 +139,19 @@ impl HttpContext for TestHttp {
 
     fn on_http_response_body(&mut self, len: usize, eof: bool) -> Action {
         info!("[hostcalls] on_response_body, {} bytes, eof: {}", len, eof);
+
+        if let Some(max) = self.body_max {
+            self.body_seen += len;
+            if self.body_seen > max {
+                self.set_http_response_body(max, 0, "".as_bytes());
+                if max == 0 {
+                    return Action::Pause;
```
thibaultcha, line 148: Ok I see now; if this is currently working then it is most likely per accident, I don't think the test cases are properly set up to catch what may be going on here. What this is doing is enabling the response buffering feature (returning `PAUSE` from `response_body`), so this feels like something is off here.

hishamhm: Thanks for the review!! I'll look at the V8 CentOS issue today then get back here to apply your suggested changes.

hishamhm: > I've had success so far with the approach I wrote in the PR review

Just to clarify: are you going to carry on implementing it or do you want me to try my hand at it? If you have a branch with those changes, even if in wip state, I can try running it on the Gateway with my modified version of the Coraza filter to check how that is behaving (or if the filter needs further changes).

thibaultcha: I am still spending some time trying to make the body work since yesterday although I got sick this week, slow progress; but I can push it later.

thibaultcha: Yesterday after the call it kind of went downhill and I could not even sit at the computer anymore. So, maybe today or maybe early next week now.

thibaultcha: Ok I just succeeded replacing the body as well with the same approach. Lots of polishing + a few more tests needed now.

thibaultcha:
Replaced by #484.

PR #471, opened 2023-12-21
chore(tests) improve hostcalls load time 

casimiro:
It's not yet clear why `parse_duration` makes Wasmtime take so long to load `hostcalls` in macOS or when instrumented by Valgrind. But by removing it, Wasmtime loads `hostcalls` significantly faster.


thibaultcha:
Excellent! Looks like it cut the Valgrind tests time in three!

PR #475, opened 2023-12-21
chore(*) update Cargo structure for Dependabot

thibaultcha:
End goal is improved integration with GitHub's security features:

1. Automated **security updates** of dependencies in our end-users Rust crates in `lib/`.
2. Automated **periodic updates** of dependencies in _all_ Rust crates in `lib/` _and_ `t/lib/` (end-users vs. test suite crates).

Changes:

- Split the single Cargo workspace we had in two: `lib/` and `t/lib`. This helps separating production vs. development crates for Dependabot.
- Minor updates to `auto/cargo` to force it to use the same `lib/ngx-wasm-rs/target/` directory as before instead of `lib/target/` with the new workspace change.
- Update `dependabot.yml` to configure several jobs for automated version bumps (distinct from security bumps).
- Clean `work/lib/wasm` on `make cleanup`.
- Always copy SDK example filters to `work/lib/wasm` due to cleanup change.
- Rename the `ngx-wasm-rs` crate purely for consistency.
- New `make update` target for quick `cargo update` in all workspaces.


auto/cargo
```diff
@@ -105,9 +111,7 @@ if [ $ngx_found = no ]; then
         exit 1
     fi
 
-    echo "$0: warning: the following functionality is not available for $ngx_addon_name:"
-    echo "    $ngx_wasm_cargo_defines"
-    echo
+    echo "$0: warning: the following functionality is not available for $ngx_addon_name: $ngx_wasm_cargo_defines"
```
thibaultcha, line 114: Not an `exit`, use same format as the above `echo "$0: error:` output.

thibaultcha:
Ping @hishamhm @casimiro for ack of operational changes in this PR.

Ping @locao and @flrgh because I believe this _may_ break the Gateway bazel builds of the Rust modules but I have not tried it yet.

thibaultcha:
Building Kong Gateway with Wasmer or V8 is already broken as of today, so we will merge this without further consideration for the Bazel build system.

lib/ngx-wasm-rs/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "ngx-wasm"
+name = "ngx-wasm-rs"
```
hishamhm, line 2: The Rust API guidelines recommend against `-rs` for the actual crate names:

https://rust-lang.github.io/api-guidelines/naming.html

> Crate names should not use -rs or -rust as a suffix or prefix. Every crate is Rust! It serves no purpose to remind users of this constantly.

So you see many projects in the wild where the repo is `foo-rs` but the crate is `foo`.

thibaultcha: Given we will not publish this create to crates.io I preferred going for consistency so it would help distinguish it from the different Rust crates we have: "`ngx-wasm-rs` is the .so lib we extra features we embed into our builds". I looked at `ngx-wasm` and its crate name is `ngx`; I'll revert this for now but a rename of both of these crates (or even a merge) is incoming in any case because all these names are more confusing than anything (also we should align with the Lua lib and name it with `wasmx` instead of `wasm`).

PR #477, opened 2024-01-03
chore(license) intact LICENSE copy and new NOTICE file

thibaultcha:
This has bugged me for some time and even though I know Kong/kong does this, I do not think it is correct and have been wanting to fix it. With a quick search I found [this article](https://vladimirgorej.com/blog/how-to-apply-apache2-license-to-your-open-source-software-project/) which I applied with the `NOTICE` file practice.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020-2023 Kong Inc.
+   Copyright [yyyy] [name of copyright owner]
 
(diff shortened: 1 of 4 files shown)


hishamhm:
Having used checksums before to track license files as a distro maintainer (which often required custom filtering because of this), I welcome this PR's practice!

I wasn't aware that Apache 2 specifically has wording for this. Interestingly, the GNU license files do not have copyright notices in them, and the copyright notices are usually put in documentation (README, etc) and/or source code, so in the (A/L)GPL world that's never an issue. I guess the "edit the copyright line" thing is a MIT/BSD license practice.

PR #484, opened 2024-01-26
feat(proxy-wasm) support 'send_local_response' during on_response_headers

thibaultcha:
Replaces #470 by implementing the feature at the Proxy-Wasm host layer.

### DIFF
--- a/src/common/proxy_wasm/ngx_proxy_wasm.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm.c
@@ -422,6 +422,10 @@ action2rc(ngx_proxy_wasm_ctx_t *pwctx,
     ngx_proxy_wasm_exec_t    *pwexecs;
     ngx_proxy_wasm_filter_t  *filter;
     ngx_proxy_wasm_action_e   action;
+#ifdef NGX_WASM_HTTP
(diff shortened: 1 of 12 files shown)


thibaultcha:
@hishamhm I recall you were running some tests with various branches for this feature, would you give it another try with this branch?

thibaultcha:
Hmmm that is strange, this should not rely to the DNS resolver at least not directly... Not sure what could be going on. I am spending some cycles on mlcache since it is being used by the new Gateway DNS resolver, but perhaps this needs some more work.

thibaultcha:
Is your filter using a dispatch call prior to the response being produced? Because if so then it can be the same bug Damon experienced last week. If so, then `dns_no_sync` alone is not enough, you will also need [this commit](https://github.com/Kong/kong/commit/4d03ca45edfa54c76e5d2d4271892f37b7524ddd). Or produce a response without a dispatch call.

thibaultcha:
@hishamhm Ok, that second commit should fix this!

hishamhm:
@thibaultcha just tested the latest version of the PR, and yes, it does pass all of the `e2e-example.sh` tests running my fork of the coraza-waf filter :tada:

PR #494, opened 2024-02-06
feat(proxy-wasm) enable multiple parallel dispatch calls

thibaultcha:
Update ngx_proxy_wasm host code to better delegate dispatch scheduling
to filters.

- Keep track of number of pending calls in ngx_proxy_wasm_ctx.
- Do not unconditionally CONTINUE after dispatch callback.
- PAUSE after dispatch callback when pending_calls > 0.


thibaultcha:
Next step: better handling responses produced from a dispatch callback. Especially when multiple callbacks are trying to produce a response: either cancel pending callbacks or improve the current "local response already stashed" error.

PR #498, opened 2024-02-08
tests(hostcalls) drop `url` dependency; improve load time

casimiro:
Removing the `url` crate, reduces `hostcalls` load time in Wasmtime by 25%.


thibaultcha:
Awesome!

PR #503, opened 2024-02-08
chore(ci) drop codeql paths/paths-ignore from its config

casimiro:
According to the GHA warning:

> The "paths"/"paths-ignore" fields of the config only have effect
for JavaScript, Python, and Ruby.

### DIFF
--- a/.github/actions/codeql-config.yml
+++ b/.github/actions/codeql-config.yml
@@ -1,6 +1,0 @@
-paths:
-  - src/
-  - lib/
-paths-ignore:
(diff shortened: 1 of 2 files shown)


thibaultcha:
We might have to reconfigure all of our CodeQL setup anyway, see https://github.com/github/codeql-action/issues/1179#issuecomment-1932932970 - I'm wondering if we should just drop it, so far all the noise is just for Perl libs in work/ and I can't find a way to make it shut up.

PR #486, opened 2024-01-29
chore(deps) fix V8 release builds

hishamhm:
Closes #485.

- [x] fix Ubuntu 18.04
- [x] fix CentOS 8
- [x] fix ArchLinux

PR #506, opened 2024-02-14
chore(ci) assign unique names to uploaded artifacts

casimiro:
Since v4, the action `upload-artifact` no longer supports uploading artifacts with non-unique names.


thibaultcha:
@casimiro Caio, would you mind drafting a PR that takes both @hishamhm's work in #486 and the bits from this PR? Basically rebasing this PR atop #486. I think `name: release-artifacts-${[ github.job }}` is a good template for all these artifacts names (combining both of your ideas). This PR also has the unit/valgrind uploads fixed and some other version bumps. Let's combine it all and enable the release job temporarily in `release.yml` to test it.

PR #507, opened 2024-02-16
chore(ci) fix release build and artifacts upload

casimiro:


thibaultcha:
#508 has fixes for the spurious [warn] issues about the socket deletion.

hishamhm:
@casimiro branch is looking good, nice to see all the release jobs in green! :green_heart: I took the liberty to push an extra tmp commit to make it list the downloaded assets, just so we can see that the addition from d74a11cab32f8cc08b18dc4f4aa228158b0e85c0 combined all assets as expected. Once we have that last bit of confirmation, I think this is good to go!

thibaultcha:
Merged in #510.

PR #520, opened 2024-03-12
fix(shm/kv) re-allocate memory if current value is replaced by a larger one

casimiro:
The first time `ngx_wasm_shm_kv_set_locked` is called with a key `k` and a value `v`, it allocates `sizeof(ngx_wasm_shm_kv_node_t) + key->len + value->len` bytes to store node, key and value contents.

If `ngx_wasm_shm_kv_set_locked` is later called with the key `k` and a value larger than `v`, it will write beyond the initially allocated memory when copying the value's bytes to the shared memory: `ngx_memcpy(n->value.data, value->data, value->len)`.


thibaultcha:
Nice!

PR #525, opened 2024-03-26
fix(ffi) allow 'load()' to fail and be invoked again

thibaultcha:
Before instantiating filter chains and request contexts, filters themselves (i.e. their root context) are represented in the module. Eventually, `ngx_proxy_wasm_start` iterates over all filters to start their root contexts.

Previously, all filters declared in `proxy_wasm` directives were stored globally. With the introduction of the FFI filters can be dynamically added to the internal representation during `proxy_wasm.load`, which makes it possible for `ngx_proxy_wasm_start` to be invoked several times.

However, since the filters structure was global then all previously declared filters were started again, including filters that previously failed initializing their root contexts. This was not happening with `proxy_wasm` directives since a filter failing initialization in that case would not allow Nginx to start in the first place.

Now all filters are stored in a "filters root" structure which can divide filter root contexts as needed. Each worker has a global "worker filters root", and each `ngx_wasm_ops_plan` created by the FFI has its own isolated "filters root".

### DIFF
--- a/src/common/lua/ngx_wasm_lua_ffi.c
+++ b/src/common/lua/ngx_wasm_lua_ffi.c
@@ -25,11 +25,12 @@ ngx_http_wasm_ffi_plan_new(ngx_wavm_t *vm,
     ngx_wasm_ffi_filter_t *filters, size_t n_filters,
     ngx_wasm_ops_plan_t **out, u_char *err, size_t *errlen)
 {
-    size_t                      i;
-    ngx_int_t                   rc;
-    ngx_wasm_ops_plan_t        *plan;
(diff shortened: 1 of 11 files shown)


t/04-openresty/ffi/200-proxy_wasm_and_lua_sanity.t
```diff
@@ -442,6 +442,7 @@ qr/.*? wasm cleaning up request pool.*
 
 
 === TEST 9: proxy_wasm FFI - filter chain attached before and during error handler location
+--- SKIP
```
thibaultcha, line 445: Still a TODO: since `next_id` is not global anymore, this can't work anymore either.

PR #532, opened 2024-04-17
feat(proxy-wasm) allow setting ":status" header

hishamhm:
Since #484, we now have an easy way to implement support for setting the ":status" pseudo-header.

### DIFF
--- a/src/common/proxy_wasm/ngx_proxy_wasm_host.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm_host.c
@@ -623,45 +623,47 @@ ngx_proxy_wasm_hfuncs_get_header_map_value(ngx_wavm_instance_t *instance,
 
 
 static ngx_int_t
-ngx_proxy_wasm_hfuncs_set_header_map_pairs(ngx_wavm_instance_t *instance,
-    wasm_val_t args[], wasm_val_t rets[])
+ngx_proxy_wasm_hfuncs_set_header_check(ngx_wavm_instance_t *instance,
+    ngx_proxy_wasm_map_type_e map_type, wasm_val_t rets[])
 {
-    ngx_int_t                         rc = NGX_ERROR;
-    ngx_proxy_wasm_map_type_e         map_type;
-    ngx_array_t                       headers;
-    ngx_proxy_wasm_marshalled_map_t   map;
-    ngx_proxy_wasm_exec_t            *pwexec;
-    ngx_proxy_wasm_ctx_t             *pwctx;
 #ifdef NGX_WASM_HTTP
-    ngx_http_wasm_req_ctx_t          *rctx;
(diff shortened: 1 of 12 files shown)


thibaultcha:
Before merging this (I have not made any changes yet locally), I think we should add context checks for `set_response_header` to better capture which phases the function can be called from (and `:status` be set from, particularly). These context checks go in `t/03-proxy_wasm/hfuncs/contexts`.

hishamhm:
> I think we should add context checks for set_response_header to better capture which phases the function can be called from (and :status be set from, particularly). 

@thibaultcha How hard should the context checks be? In the current code, headers being set in the wrong context are soft errors, and bodies being set in the wrong context are hard errors. 

For headers, functions such as `ngx_proxy_wasm_hfuncs_replace_header_map_value` and `ngx_proxy_wasm_hfuncs_set_header_map_pairs` return OK and just log an error when attempting to set response header when the headers are already sent.

For bodies, in `ngx_proxy_wasm_hfuncs_set_buffer`, attempting to set the response body or request body from the wrong place trigger a trap.

To clarify, by "add context checks" are you suggesting we turn the header checks mentioned above into hard errors?

hishamhm:
@thibaultcha is this second commit [6014ff0](https://github.com/Kong/ngx_wasm_module/pull/532/commits/6014ff0e626282adff2415cf1a95d5459f09e503) what you had in mind?

thibaultcha:
> @thibaultcha is this second commit [6014ff0](https://github.com/Kong/ngx_wasm_module/pull/532/commits/6014ff0e626282adff2415cf1a95d5459f09e503) what you had in mind?

Yes, strengthen our context checks and formalize the behavior in each unsupported phase. Regarding the tests themselves, setting a response header in the request phases will not have any effect so I think we should catch usage there is what I meant, cutting some decisions where the behavior is "strange". In other words, solidify the response header behavior especially if we introduce `:status`, which *may* work when used from the request phases(?). As we add more unknowns into the response headers, I think we should cover all expected behaviors and make the decisions wherever we have margin.

hishamhm:
@thibaultcha The added context checks and tests should be ready for review now!

thibaultcha:
Thanks! The code etc all looks good; I'll just reorganize the tests to try not to duplicate test cases between `hfuncs/` and `hfuncs/contexts` (also the contexts tests are supposed to be non-functional - only test the return code - but maybe they should be to avoid confusion, I'll compare with the other suites).

hishamhm:
Thanks! I did notice some duplication there, but when in doubt I added stuff to both, just in case. The main difference I could spot is that `hfuncs/contexts` tests are using `context-checks` which is testing at the `extern "C"` low-level (e.g `proxy_set_buffer_bytes`), and the ones in `hfuncs/` are using proxy-wasm-rust-sdk (e.g. `ctx.set_http_response_headers`).

PR #536, opened 2024-04-23
feat(wasmtime) expose cache_config directive

hishamhm:
This adds support for `wasmtime { flag cache_config_load <config.toml>; }`, for enabling disk caching of ahead-of-time compilation of .wasm files.

This should alleviate noticeable CPU spikes when starting up workers using Wasmtime and Wasm modules larger than a few megabytes.

Implementation notes:

* Added error checking to the config directive, even though not all wasmtime config checks do, because this one reports various errors such as file not found and parse errors in the config file.
* Kept the same name `cache_config_load` as documented in https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.cache_config_load since that is the reference we point to in our documentation.

### DIFF
--- a/src/wasm/wrt/ngx_wrt_wasmtime.c
+++ b/src/wasm/wrt/ngx_wrt_wasmtime.c
@@ -17,6 +17,10 @@ typedef void (*wasmtime_config_set_int_pt)(wasm_config_t *config,
 typedef void (*wasmtime_config_set_bool_pt)(wasm_config_t *config, bool value);
 
 
+static u_char * ngx_wasmtime_log_handler(ngx_wrt_res_t *res, u_char *buf,
+    size_t len);
+
+
 static ngx_int_t
 size_flag_handler(wasm_config_t *config, ngx_str_t *name, ngx_str_t *value,
     ngx_log_t *log, void *wrt_setter)
@@ -129,7 +133,10 @@ profiler_flag_handler(wasm_config_t *config, ngx_str_t *name, ngx_str_t *value,
 static wasm_config_t *
 ngx_wasmtime_init_conf(ngx_wavm_conf_t *conf, ngx_log_t *log)
 {
-    wasm_config_t  *config;
+    wasm_config_t     *config;
+    char              *pathname;
+    u_char            *errmsg;
+    wasmtime_error_t  *err;
 #if 0
     wasm_name_t        msg;
     wasmtime_error_t  *err = NULL;
@@ -159,6 +166,39 @@ ngx_wasmtime_init_conf(ngx_wavm_conf_t *conf, ngx_log_t *log)
     wasmtime_config_cranelift_opt_level_set(config, WASMTIME_OPT_LEVEL_NONE);
     wasmtime_config_static_memory_maximum_size_set(config, 0);
 #endif
+
+    if (conf->cache_config.len) {
+        ngx_wavm_log_error(NGX_LOG_INFO, log, NULL,
+                           "setting wasmtime cache config file: \"%V\"",
+                           &conf->cache_config);
+
+        pathname = ngx_calloc(conf->cache_config.len + 1, log);
+        if (pathname == NULL) {
+            goto error;
+        }
+
+        ngx_memcpy(pathname, conf->cache_config.data, conf->cache_config.len);
+
+        err = wasmtime_config_cache_config_load(config, pathname);
+
+        ngx_free(pathname);
+
+        if (err) {
+            errmsg = ngx_calloc(NGX_MAX_ERROR_STR + 1, log);
+            if (errmsg == NULL) {
+                goto error;
+            }
+
+            ngx_wasmtime_log_handler((ngx_wrt_res_t *) err, errmsg,
+                                     NGX_MAX_ERROR_STR);
+
+            ngx_log_error(NGX_LOG_EMERG, log, 0,
+                          "failed configuring wasmtime cache; %s",
+                          errmsg);
+
+            goto error;
+        }
+    }
 
     if (conf->compiler.len) {
         if (ngx_str_eq(conf->compiler.data, conf->compiler.len,
--- a/t/01-wasm/directives/011-wasmtime_cache_config_directive.t
+++ b/t/01-wasm/directives/011-wasmtime_cache_config_directive.t
@@ -1,0 +1,124 @@
+# vim:set ft= ts=4 sts=4 sw=4 et fdm=marker:
+
+use strict;
+use lib '.';
+use t::TestWasmX;
+
+our $nginxV = $t::TestWasmX::nginxV;
+our $osname = $t::TestWasmX::osname;
+
+add_cleanup_handler(sub {
+    my @jit_dumps = glob("$::pwd/jit-*.dump");
+
+    foreach my $file (@jit_dumps) {
+        #warn "\n", $file, "\n";
+        unlink($file);
+    }
+});
+
+plan_tests(4);
+run_tests();
+
+__DATA__
+
+=== TEST 1: wasmtime cache_config - missing file
+--- skip_eval: 4: $::nginxV !~ m/wasmtime/
+--- main_config
+    wasm {
+        wasmtime {
+            cache_config missing_file;
+        }
+    }
+--- error_log eval
+qr/\[emerg\] .*? failed configuring wasmtime cache; failed to read config file: missing_file/
+--- no_error_log
+[error]
+[crit]
+--- must_die
+
+
+
+=== TEST 2: wasmtime cache_config - bad file
+--- skip_eval: 4: $::nginxV !~ m/wasmtime/
+--- user_files
+>>> wasmtime_config.toml
+invalid contents
+--- main_config
+    wasm {
+        wasmtime {
+            cache_config $TEST_NGINX_HTML_DIR/wasmtime_config.toml;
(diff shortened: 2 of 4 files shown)


hishamhm:
@thibaultcha the tests for cache enabled and disabled don't actually check that the cache was actually created (I don't know how to check for the existence of files at the end of a test within the Test::Nginx DSL), but if you run each of them with `-- ONLY` you can see that one of them creates the cache in the temporary dir and the other one doesn't.

thibaultcha:
Ultimately it's fine if all we test is our directive and its calling of the underlying wasmtime function. We aren't responsible for testing the behavior of the runtime itself, even though that's a plus if we can, but well...

t/01-wasm/directives/008-flag_directive_wasmtime.t
```diff
@@ -87,7 +87,111 @@ qr/setting flag: "debug_info=off"/
 
 
 
-=== TEST 5: flag directive - wasmtime - consume_fuel - on
+=== TEST 5: flag directive - wasmtime - cache_config_load - missing file
+--- skip_eval: 4: $::nginxV !~ m/wasmtime/
+--- main_config
+    wasm {
+        wasmtime {
+            flag cache_config_load missing_file;
```
thibaultcha, line 95: Should we implement this a level above the `flag` directive? It reads a little weird or is it just me? E.g:
```
wasmtime {
    cache_config  /path/to/file;
}
```

flrgh: Does wasmtime expose an API for setting its cache parameters directly? It's not the end of the world if we have to generate the toml file and pass it in, but something like this would be a lot more ergonomic:

```
wasmtime {
  # implies cache.enabled = true
  cache_directory = /path/to/dir;
}
```

hishamhm: It does read a bit weird — I originally called it just `cache_config`, but then when I checked our docs to document it, the docs pointed to the Wasmtime Config struct page, and all of our flag names were 1-to-1 with the method calls for that struct, so I aligned it to that method call name for consistency.

The other top-level entries we have tend to be shared across VMs (`backtraces`, `compiler`) and this one is very Wasmtime-specific (it would end up declared at `ngx_wasm_core_module.c`?), so maybe keeping it under `flag` is fine...

As for the name, I wouldn't mind calling it `flag cache_config /path/to/file` and then specifically documenting it in `docs/DIRECTIVES.md`.

hishamhm: @flrgh [it doesn't](https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.cache_config_load), and the actual .toml has [a lot of options](https://docs.wasmtime.dev/cli-cache.html), so it's probably for the best if you have the flexibility to write them in the .toml on your end rather than relying on which options are exposed or not by wasmx given a particular version of wasmtime...

thibaultcha: It could still be wasmtime specific but not part of `flag` since we have the `wasmtime{}` block for the purpose of runtime-specific configuration directives. All `flag` handlers are supposed to be generic (e.g. `bool_flag_handler` or `size_flag_handler`) and receive a `wrt_setter` pointer, which is unused in the current implementation of `cache_config_load_flag_handler`. We could have `string_flag_handler` and give the `wasmtime_config_cache_config_load` as the `wrt_setter`, if that works (didn't check).

hishamhm: @thibaultcha I attempted to implement it as a `wasmtime`-level `config_cache` directive in this branch: https://github.com/Kong/ngx_wasm_module/tree/feat/wasmtime-config-cache-no-flag

...but it's giving me `invalid number of arguments in "cache_config"` errors, and I can't figure out why. I used the same NGX_CONF_TAKE1 flag as the `"compiler"` directive right below it.

Any idea why?

thibaultcha: It looks like an issue with our definitions :/ We should hit [this](https://github.com/nginx/nginx/blob/master/src/core/ngx_conf_file.c#L419) expression successfully and pass the arguments check, but we don't pass [these two](https://github.com/nginx/nginx/blob/master/src/core/ngx_conf_file.c#L415-L417) previous conditions (I gdb'ed it).

thibaultcha: This took care of it:
```patch
diff --git a/src/wasm/ngx_wasm.h b/src/wasm/ngx_wasm.h
index d832f026..c570d10b 100644
--- a/src/wasm/ngx_wasm.h
+++ b/src/wasm/ngx_wasm.h
@@ -14,10 +14,10 @@
 #define NGX_LOG_WASM_NYI             NGX_LOG_ALERT
 
 #define NGX_WASM_MODULE              0x5741534d   /* "WASM" */
-#define NGX_WASM_CONF                0x00300000
-#define NGX_WASMTIME_CONF            0x00001000
-#define NGX_WASMER_CONF              0x00002000
-#define NGX_V8_CONF                  0x00004000
+#define NGX_WASM_CONF                0x10000000
+#define NGX_WASMTIME_CONF            0x20000000
+#define NGX_WASMER_CONF              0x40000000
+#define NGX_V8_CONF                  0x80000000
 
 #define NGX_WASM_DONE_PHASE          15
 #define NGX_WASM_BACKGROUND_PHASE    16
diff --git a/src/wasm/wrt/ngx_wrt_wasmtime.c b/src/wasm/wrt/ngx_wrt_wasmtime.c
index 802ab5ff..2ed0f1c2 100644
--- a/src/wasm/wrt/ngx_wrt_wasmtime.c
+++ b/src/wasm/wrt/ngx_wrt_wasmtime.c
@@ -171,7 +171,7 @@ ngx_wasmtime_init_conf(ngx_wavm_conf_t *conf, ngx_log_t *log)
     if (conf->cache_config.len) {
         ngx_wavm_log_error(NGX_LOG_INFO, log, NULL,
                            "setting wasmtime cache config file: \"%V\"",
-                           conf->cache_config);
+                           &conf->cache_config);
 
         pathname = ngx_calloc(conf->cache_config.len + 1, log);
         if (pathname == NULL) {
```

I hope the `NGX_WASM_CONF` change works for the whole suite.

hishamhm: Thanks! Rebased.

src/wasm/wrt/ngx_wrt_wasmtime.c
```diff
@@ -8,15 +8,20 @@
 #include <ngx_wasi.h>
 
 
-#define NGX_WRT_WASMTIME_STACK_NARGS  10
-#define NGX_WRT_WASMTIME_STACK_NRETS  10
+#define NGX_WRT_WASMTIME_STACK_NARGS     10
+#define NGX_WRT_WASMTIME_STACK_NRETS     10
+#define NGX_WRT_WASMTIME_CONFIG_ERR_LEN  255
```
thibaultcha, line 13: Could we use `NGX_MAX_ERROR_STR` and avoid a new definition?

t/01-wasm/directives/008-flag_directive_wasmtime.t
```diff
@@ -87,7 +87,111 @@ qr/setting flag: "debug_info=off"/
 
 
 
-=== TEST 5: flag directive - wasmtime - consume_fuel - on
+=== TEST 5: flag directive - wasmtime - cache_config_load - missing file
```
thibaultcha, line 90:
```suggestion
=== TEST 5: wasmtime cache_config - missing file
```

We can also rename the file name with `008-wasmtime_cache_config.t`

hishamhm: Oops, good catch. 

> We can also rename the file name with 008-wasmtime_cache_config.t

008-flag_directive_wasmtime.t should stay as is since it has the rest of the wasmtime flag tests.  I added a new file for this directive, like we do for the other standalone directives.

thibaultcha: Didn't realize it was the same file; yes for a new one.

t/01-wasm/directives/011-wasmtime_cache_config_directive.t
```diff
@@ -0,0 +1,124 @@
...
+our $osname = $t::TestWasmX::osname;
+
+add_cleanup_handler(sub {
+    my @jit_dumps = glob("$::pwd/jit-*.dump");
+
+    foreach my $file (@jit_dumps) {
+        #warn "\n", $file, "\n";
+        unlink($file);
+    }
+});
```
thibaultcha, line 17: Do we need this in this file?

t/01-wasm/directives/011-wasmtime_cache_config_directive.t
```diff
@@ -0,0 +1,124 @@
+# vim:set ft= ts=4 sts=4 sw=4 et fdm=marker:
+
+use strict;
+use lib '.';
+use t::TestWasmX;
+
+our $nginxV = $t::TestWasmX::nginxV;
+our $osname = $t::TestWasmX::osname;
```
thibaultcha, line 8: Also looks unused

PR #537, opened 2024-04-24
[merge] feat(proxy-wasm) allow setting ":status" header

thibaultcha:


src/common/proxy_wasm/ngx_proxy_wasm_maps.c
```diff
@@ -702,6 +705,39 @@ ngx_proxy_wasm_maps_get_response_status(ngx_wavm_instance_t *instance,
...
+         * Ideally we would like to return NGX_ERROR here, but the
+         * proxy-wasm-rust-sdk causes a panic when the host returns an error
+         * code. So we ignore invalid values here. This allows the application
+         * code to gracefully check success by setting then getting.
+         */
+        pwexec = ngx_proxy_wasm_instance2pwexec(instance);
+
+        ngx_proxy_wasm_log_error(NGX_LOG_WARN, pwexec->log, 0,
+                                 "ignoring attempt to set invalid response "
+                                 "status: \"%V\"", value);
```
thibaultcha, line 731: @hishamhm Here I narrowed the status range as per Nginx does in other places, and used the same error message for the template. I think a warning here is better than nothing?

hishamhm: sounds good to me!

Issue #528, opened 2024-04-09
Memory corruption when dispatch socket outlives request


hishamhm:
I've been observing frequent segfaults when running the [datakit](https://github.com/kong/datakit-filter) filter on a `kong/kong:nightly` container which uses ngx_wasm_module commit d5e4c698c51a6ef39551184f99eb4d1605f671d9 and a filter configuration that dispatches multiple HTTP subrequests. (I could not reproduce the problem on a configuration with a single HTTP dispatch).

This was one of the segfaults:

<details>
```
Program received signal SIGSEGV, Segmentation fault.
0x000057f6f83fade1 in ngx_palloc_large (pool=pool@entry=0x57f6fa0e2710, size=size@entry=1120) at src/core/ngx_palloc.c:244
244	src/core/ngx_palloc.c: No such file or directory.
(gdb) bt
#0  0x000057f6f83fade1 in ngx_palloc_large (pool=pool@entry=0x57f6fa0e2710, size=size@entry=1120) at src/core/ngx_palloc.c:244
#1  0x000057f6f83fafdf in ngx_palloc (pool=pool@entry=0x57f6fa0e2710, size=size@entry=1120) at src/core/ngx_palloc.c:131
#2  0x00007723db389217 in ngx_list_init (size=56, n=20, pool=0x57f6fa0e2710, list=0x57f6f9f32ec0) at src/core/ngx_list.h:39
#3  ngx_wasm_socket_read_http_response (sock=0x57f6f9f328e8, bytes=539, ctx=0x57f6f9f32cc8)
    at /home/runner/.cache/bazel/_bazel_runner/5f5cb99aeb7e2b4db19484ea5022e949/execroot/kong/external/ngx_wasm_module/src/common/ngx_wasm_socket_tcp.c:968
#4  0x00007723db3894ba in ngx_wasm_socket_tcp_read (sock=sock@entry=0x57f6f9f328e8, reader=0x7723db389150 <ngx_wasm_socket_read_http_response>, 
    reader_ctx=reader_ctx@entry=0x57f6f9f32cc8)
    at /home/runner/.cache/bazel/_bazel_runner/5f5cb99aeb7e2b4db19484ea5022e949/execroot/kong/external/ngx_wasm_module/src/common/ngx_wasm_socket_tcp.c:1064
#5  0x00007723db39b934 in ngx_http_proxy_wasm_dispatch_resume_handler (sock=<optimized out>)
[...]

PR #545, opened 2024-05-06
tests(proxy_wasm) check parallel calls with failing connections

hishamhm:
This is a minimal reproducer to a segfault I originally triggered with Datakit, where you can get a crash when performing multiple dispatch calls with failing connections. The first dispatch is terminated with `dispatch failed: tcp socket - Connection refused`, and this seems to free resources that are still being used by the second dispatch, leading to [memory corruption and a segfault](https://github.com/Kong/ngx_wasm_module/actions/runs/8972623154/job/24641007935).

This looks similar to #528 (both include multiple dispatches and the Valgrind errors refer to accessing data freed by `ngx_http_finalize_connection`, etc.; both might have the same root cause), but in [that case](https://github.com/Kong/ngx_wasm_module/files/14924491/full_log.txt) I didn't get `dispatch failed: tcp socket` in the full log.

### DIFF
--- a/t/03-proxy_wasm/hfuncs/133-proxy_dispatch_http_edge_cases.t
+++ b/t/03-proxy_wasm/hfuncs/133-proxy_dispatch_http_edge_cases.t
@@ -373,3 +373,26 @@ qr/\A.*? on_request_headers.*
 [error]
 [crit]
 [emerg]
+
+
+
+=== TEST 11: proxy_wasm - dispatch_http_call() 2 failing parallel calls
+--- valgrind
+--- load_nginx_modules: ngx_http_echo_module
+--- wasm_modules: hostcalls
+--- config
+    location /t {
+        proxy_wasm hostcalls 'on=request_headers \
+                              test=/t/dispatch_http_call \
+                              host=127.0.0.1:1 \
+                              path=/bad_port \
+                              ncalls=2';
+        echo ok;
+    }
+--- error_code: 500
+--- no_error_log
+[crit]
+[emerg]
+[alert]
+[stub1]
+[stub2]


thibaultcha:
I spent some time on this, the fix is clear but I don't think I can have it ready by tomorrow. I initially thought it would be doable so I wrote it on main, but it is larger than I expected. In the fix I changed the dispatch behavior to not stop the proxy_wasm runloop as before, and let an individual dispatch fail while others are still running (and may succeed). This needs more tests which I haven't written, but it also causes a couple new failures in the FFI which makes me want to write the fix against the Lua bridge refactor instead at this point, especially if a smaller patch by tomorrow isn't possible...

hishamhm:
> In the fix I changed the dispatch behavior to not stop the proxy_wasm runloop as before on error, and let an individual dispatch fail while others are still running (and may succeed). We can't really return 500 after only one call failed, that isn't very consistent. We also can't return 500 for "all dispatch failed", that's pretty hard to try and also very arbitrary... I may need to check Envoy behavior. 

Yes, regardless of this bug I was going to discuss this behavior with you as well. In principle, failing a dispatch shouldn't trigger a failure, it's something that I'd definitely want to catch and handle in Datakit (one could think of flows that go, "try this API URL, if that fails try this other one", etc.)

thibaultcha:
Replaced with #546. @hishamhm I would like for the datakit filter to be tested with #546 before merging it, but the FFI breaking change must also be applied to the Gateway (`proxy_wasm.start()` calls are to be removed, but that should be all). cc @flrgh for heads-up on the upcoming FFI change as well.

hishamhm:
> @hishamhm I would like for the datakit filter to be tested with https://github.com/Kong/ngx_wasm_module/pull/546 before merging it

I've just tested it and it's still crashing. Running it enough times I do get segfaults, but every run is reporting memory access errors consistently.

These are the logs from a Gateway run with Valgrind. This particular run didn't crash, but Valgrind reported reads on data that was freed by `ngx_proxy_wasm_dispatch_calls_cancel`:
[pr-546-kong-error.log](https://github.com/Kong/ngx_wasm_module/files/15351164/pr-546-kong-error.log)
[pr-546-kong-valgrind.log](https://github.com/Kong/ngx_wasm_module/files/15351167/pr-546-kong-valgrind.log)

This was running a datakit configuration with 2 parallel calls that fail (I didn't spin up the other localhost upstream).

PR #554, opened 2024-06-19
[merge] feat/metrics

thibaultcha:
- ~[ ] `ngx_wa_*` facilities getters~
- [x] default slab?

### DIFF
--- a/src/common/metrics/ngx_wa_metrics.c
+++ b/src/common/metrics/ngx_wa_metrics.c
@@ -1,0 +1,504 @@
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+#include <ngx_wasm.h>
+#include <ngx_wa_metrics.h>
+#include <ngx_wa_histogram.h>
+
+
+ngx_str_t *
+ngx_wa_metric_type_name(ngx_wa_metric_type_e type)
+{
+    static ngx_str_t  counter = ngx_string("counter");
+    static ngx_str_t  gauge = ngx_string("gauge");
+    static ngx_str_t  histogram = ngx_string("histogram");
+    static ngx_str_t  unknown = ngx_string("unknown");
+
+    switch (type) {
+    case NGX_WA_METRIC_COUNTER:
+        return &counter;
+
+    case NGX_WA_METRIC_GAUGE:
+        return &gauge;
+
+    case NGX_WA_METRIC_HISTOGRAM:
+        return &histogram;
+
+    default:
+        return &unknown;
+    }
+}
+
+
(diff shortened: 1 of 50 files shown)


src/common/metrics/ngx_wa_metrics.c
```diff
@@ -0,0 +1,460 @@
...
+
+    return rc;
+}
+
+
+ngx_int_t
+ngx_wa_metrics_increment(ngx_wa_metrics_t *metrics, uint32_t mid, ngx_int_t n)
+{
+    uint32_t          cas;
+    ngx_uint_t        slot;
```
casimiro, line 329: nitpick: maybe declare the variable `slot` after `val` so that variable declarations are sorted by the length of the type name.

thibaultcha: Length of type name isn't really the sorting criteria

thibaultcha: In this specific case, I just didn't like that a stack value was in between 2 pointer values, so I moved it up along the other unsigned integer variable. Sometimes re-arranging by length is more aesthetic but in this case I think this is more neat.

thibaultcha:
@casimiro I added one more fix that I can squash later, can you take a quick look at it make sure it is sound? I tested the `ngx_init_cycle` code paths locally.

casimiro:
Nice catch @thibaultcha!

The added change indeed fixes the issue, although I think there's a more terse way to fix it.

Originally, we were setting `metrics->shm_zone->noreuse` to 1 in `ngx_wa_metrics_init_conf` when the old_cycle's metrics couldn't be reused. Then, later in `ngx_wa_metrics_init` we would check for `metrics->shm_zone->noreuse` and reallocate metrics accordingly.

However, we should have simply set `metrics->old_metrics->shm_zone->noreuse` to 1 in `ngx_wa_metrics_init_conf` and check for `metrics->old_metrics->shm_zone->noreuse` in `ngx_wa_metrics_init`, though.

This expresses more clearly what's happening: the new configuration made the old-cycle's metric non-reusable, not the current-cycle's one. 

I pushed a commit representing what I mean.

casimiro:
I thought the commit I pushed was equivalent to the fix you proposed, weird. I'm investigating why it's not.

casimiro:
After correcting the fix I proposed, I'm less inclined to prefer it over the one you pushed.

thibaultcha:
Well are mis-using the `noreuse` flag in any case so it may not do what we want it to do. At least the original fix fully leans into that misuse and takes advantage of that it's possible to set the flag at reload (although a misuse).

PR #559, opened 2024-06-26
chore(deps) bump proxy-wasm-go-sdk to 0.24.0

thibaultcha:


thibaultcha:
This is failing on CI but my distribution is still on tinygo 0.31.2, so this will be pending for now.

PR #565, opened 2024-07-10
fix(proxy-wasm) only inject shim headers after on_response_headers

thibaultcha:
Due to some Nginx internals (some response headers are added after our module is executed but before the response is sent) we inject "shim headers" into the `get_http_response_headers()` API. Some heuristics are used to determine whether or not `Transfer-Encoding`, `Connection`, or other "last minute" headers will be injected or not.

The issue is that the heuristics used are only valid during response production and after (e.g. `r->chunked` or
`r->headers_out.content_length_n` may only be relevant during and after response production). If a user uses `get_http_response_headers()` in request phases, the returned headers may include elements that will not actually be included in the response (e.g. `Transfer-Encoding`).

We now enforce that shim response headers should only be injected during and after `on_http_response_headers`.

### DIFF
--- a/src/common/proxy_wasm/ngx_proxy_wasm_maps.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm_maps.c
@@ -143,6 +143,8 @@ ngx_proxy_wasm_maps_get_all(ngx_wavm_instance_t *instance,
 #ifdef NGX_WASM_HTTP
     ngx_table_elt_t            *shim;
     ngx_array_t                *shims;
+    ngx_proxy_wasm_ctx_t       *pwctx;
+    ngx_proxy_wasm_exec_t      *pwexec;
 #endif
 
     list = ngx_proxy_wasm_maps_get_map(instance, map_type);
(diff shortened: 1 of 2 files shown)


src/common/proxy_wasm/ngx_proxy_wasm_maps.c
```diff
@@ -178,9 +180,18 @@ ngx_proxy_wasm_maps_get_all(ngx_wavm_instance_t *instance,
...
-        if (map_type == NGX_PROXY_WASM_MAP_HTTP_RESPONSE_HEADERS) {
-            /* inject shim response headers
-             * (produced by ngx_http_header_filter)
+        pwexec = ngx_proxy_wasm_instance2pwexec(instance);
+        pwctx = pwexec->parent;
+
+        if (map_type == NGX_PROXY_WASM_MAP_HTTP_RESPONSE_HEADERS
+            && (pwctx->step == NGX_PROXY_WASM_STEP_RESP_HEADERS
+                || pwctx->step == NGX_PROXY_WASM_STEP_RESP_BODY
+                || pwctx->step == NGX_PROXY_WASM_STEP_LOG))
```
thibaultcha, line 189: Could be `pwctx->step >= NGX_PROXY_WASM_STEP_RESP_HEADERS`

PR #582, opened 2024-08-08
feat(ffi) expose kv & metrics shms through FFI

casimiro:


src/common/shm/ngx_wasm_shm.c
```diff
@@ -54,6 +54,13 @@ ngx_wasm_shm_init(ngx_cycle_t *cycle)
             }
             dd("queue init done");
             break;
+        case NGX_WASM_SHM_TYPE_METRICS:
+            rc = ngx_wasm_shm_kv_init(shm);
```
thibaultcha, line 58: This is also called in `ngx_wa_metrics_init` - can we only keep one invocation? Or perhaps better, invoke `ngx_wa_metrics_init` here instead of in `ngx_wasmx.c`?

casimiro: Actually, this PR was moving `ngx_wasm_shm_kv_init` out of `ngx_wa_metrics_init` to here. 

I agree that it makes more sense to have `ngx_wa_metrics_init` in here, rather than in `ngx_wasmx.c`, and put `ngx_wasm_shm_kv_init` back to `ngx_wa_metric_init`, though.

thibaultcha: I missed the line that removed the old `ngx_wasm_shm_kv_init`. But yes, let's do the other thing which I think is more explicit to readers of this `ngx_wasm_shm_init` switch case.

src/ngx_wasmx.h
```diff
@@ -3,6 +3,7 @@
 
 
 #include <ngx_core.h>
+#include <ngx_wasm_shm.h>
```
thibaultcha, line 6: If use this header to the wasmx module, then it needs to be moved the the wasmx module sources in `config` (and become `ngx_wa_shm.h,c`) - the wasmx module can be built for IPC only today, so no `ngx_wasm` symbols would be available.

thibaultcha: Oh, it already is there... My bad! So this is part of the rename leftover, no worries.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+        error("n must be a positive number", 2)
+    end
+
+    local shm = zone[WASM_SHM_KEY]
+    local nkeys = C.ngx_wa_ffi_shm_get_keys(shm, 0, nil)
+    if nkeys == FFI_ERROR then
+        return nil, "failed retrieving shm zone keys total"
+    end
+
+    nkeys = (max and nkeys > max) and max or nkeys
```
thibaultcha, line 150: Not sure we are keeping this (why can't we just pass `max` to the FFI call?), but this looks like `math.min`.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+
+
+function _M.init()
+    if initialized then
+        return
+    end
+
+    local rc = C.ngx_wa_ffi_shm_get_zones(get_zones_handler)
+    if rc == FFI_DECLINED then
+        ngx_log(ngx.INFO, "no shm zones found")
```
thibaultcha, line 461: Maybe a debug log at most? This log will also not have any context so we might want to add some, something like "no shm zones found for resty.wasmx interface"

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+]]
+
+
+local function get_keys(zone, max)
+    if max ~= nil and (type(max) ~= "number" or max < 1) then
+        error("n must be a positive number", 2)
+    end
+
+    local shm = zone[WASM_SHM_KEY]
+    local nkeys = C.ngx_wa_ffi_shm_get_keys(shm, 0, nil)
```
thibaultcha, line 145: Doesn't this call sort of voids the benefit of the `max` argument? This means we retrieve all keys (with a lock) always, just to find out the total number of keys and retrieve that exact number again? What am I missing?

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+    nkeys = (max and nkeys > max) and max or nkeys
+
+    local ckeys = ffi_new("ngx_str_t *[?]", nkeys)
+    local pckeys = ffi_cast("ngx_str_t **", ckeys)
+    local rc = C.ngx_wa_ffi_shm_get_keys(shm, nkeys, pckeys)
+    if rc == FFI_ERROR then
+        return nil, "failed retrieving shm zone keys"
+    end
+
+    local keys = table_new(0, nkeys)
```
thibaultcha, line 159: If we make `ngx_wa_ffi_shm_get_keys` return the number of keys in `pckeys`, wouldn't that let us avoid having to retrieve the total `nkeys` of the shm in the first place?

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+
+local function record_metric(zone, metric_id, value)
+    if type(metric_id) ~= "number" then
+        error("metric_id must be a number", 2)
+    end
+
+    if type(value) ~= "number" then
+        error("value must be a number", 2)
+    end
+
```
thibaultcha, line 383:
```suggestion
```

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+    if value ~= nil then
+        if type(value) ~= "number" or value < 1 then
+            error("value must be greater than zero", 2)
+        end
+
+        increment = value
+    end
+
```
thibaultcha, line 412:
```suggestion
```

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+local C = ffi.C
+local error = error
+local type = type
+local table_new = table.new
+local ffi_cast = ffi.cast
+local ffi_fill = ffi.fill
+local ffi_new = ffi.new
+local ffi_str = ffi.string
+local ngx_log = ngx.log
+local initialized = false
```
thibaultcha, line 16: Let's move `initialized` and `WASM_SHM_KEY` together away from the deref declarations.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,471 @@
...
+
+local function define_metric(zone, name, metric_type)
+    if type(name) ~= "string" or name == "" then
+        error("name must be a non-empty string", 2)
+    end
+
+    if metric_type_set[metric_type] == nil then
+        error("metric_type must be either \
+               resty.wasmx.shm.metrics.COUNTER, \
+               resty.wasmx.shm.metrics.GAUGE or \
```
thibaultcha, line 360:
```suggestion
               resty.wasmx.shm.metrics.GAUGE, or \
```

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -256,4 +257,173 @@ ngx_http_wasm_ffi_set_host_properties_handlers(ngx_http_request_t *r,
...
+
+ngx_int_t
+ngx_wa_ffi_shm_get_zones(ngx_wa_ffi_shm_get_zones_handler_pt handler)
+{
+    ngx_uint_t             i;
+    ngx_array_t           *shms = ngx_wasmx_shms((ngx_cycle_t *) ngx_cycle);
+    ngx_wa_shm_t          *shm;
+    ngx_wa_shm_mapping_t  *mappings;
+
+    if (!handler || !shms) {
```
thibaultcha, line 270: `== NULL` for both

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -256,4 +257,173 @@ ngx_http_wasm_ffi_set_host_properties_handlers(ngx_http_request_t *r,
...
+}
+
+
+static void
+shm_kv_retrieve_keys(ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel,
+                     ngx_int_t *total, ngx_int_t limit, ngx_str_t **keys)
+{
+    ngx_wa_shm_kv_node_t *n = (ngx_wa_shm_kv_node_t *) node;
+
+    if (!node || node == sentinel || (limit > 0 && *total == limit)) {
```
thibaultcha, line 291: `node == NULL`

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -256,4 +257,173 @@ ngx_http_wasm_ffi_set_host_properties_handlers(ngx_http_request_t *r,
 
     return ngx_proxy_wasm_properties_set_ffi_handlers(pwctx, getter, setter, r);
 }
+
+
+ngx_int_t
+ngx_wa_ffi_shm_get_zones(ngx_wa_ffi_shm_get_zones_handler_pt handler)
```
thibaultcha, line 263:
```suggestion
ngx_wa_ffi_shm_setup_zones(ngx_wa_ffi_shm_get_zones_handler_pt setup_handler)
```

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -256,4 +257,173 @@ ngx_http_wasm_ffi_set_host_properties_handlers(ngx_http_request_t *r,
...
+}
+
+
+ngx_int_t
+ngx_wa_ffi_shm_get_keys(ngx_wa_shm_t *shm, ngx_uint_t n, ngx_str_t **keys)
+{
+    ngx_int_t         total = 0;
+    ngx_wa_shm_kv_t  *kv;
+
+    if (!shm || shm->type == NGX_WASM_SHM_TYPE_QUEUE) {
```
thibaultcha, line 312: `shm == NULL` (ditto for all NULL comparison with `!` in the PR - `!` is only for numbers in our convention).

src/common/metrics/ngx_wa_metrics.h
```diff
@@ -74,7 +88,25 @@ ngx_int_t ngx_wa_metrics_increment(ngx_wa_metrics_t *metrics,
...
+static ngx_inline ngx_wa_metrics_histogram_t *
+ngx_wa_metrics_set_histogram_buffer(ngx_wa_metric_t *m, u_char *b, size_t s)
+{
+    m->slots[0].histogram = (ngx_wa_metrics_histogram_t *) b;
+    m->slots[0].histogram->n_bins = (s - sizeof(ngx_wa_metrics_histogram_t))
+                                    / sizeof(ngx_wa_metrics_bin_t);
+    m->slots[0].histogram->bins[0].upper_bound = NGX_MAX_UINT32_VALUE;
+
+    return m->slots[0].histogram;
+}
```
thibaultcha, line 110:
```suggestion
}

```

src/common/metrics/ngx_wa_metrics.h
```diff
@@ -74,7 +88,25 @@ ngx_int_t ngx_wa_metrics_increment(ngx_wa_metrics_t *metrics,
...
+ngx_int_t ngx_wa_metrics_histogram_add_locked(ngx_wa_metrics_t *metrics,
+    ngx_wa_metric_t *m);
+ngx_int_t ngx_wa_metrics_histogram_record(ngx_wa_metrics_t *metrics,
+    ngx_wa_metric_t *m, ngx_uint_t slot, uint32_t mid, ngx_uint_t n);
+void ngx_wa_metrics_histogram_get(ngx_wa_metrics_t *metrics, ngx_wa_metric_t *m,
+    ngx_uint_t slots, ngx_wa_metrics_histogram_t *out);
+
+
+static ngx_inline ngx_wa_metrics_histogram_t *
+ngx_wa_metrics_set_histogram_buffer(ngx_wa_metric_t *m, u_char *b, size_t s)
```
thibaultcha, line 102:
```suggestion
ngx_wa_metrics_histogram_set_buffer(ngx_wa_metric_t *m, u_char *b, size_t s)
```

lib/resty/wasmx/shm.lua
casimiro: Good call.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,480 @@
...
+        end
+
+        total = tonumber(ctotal[0])
+
+        if total > 0 then
+            local keys = table_new(0, total)
+
+            for i = 1, total do
+                local ckey = ckeys[i-1][0]
+                local key = ffi_str(ckey.data, ckey.len)
```
casimiro, line 169: This local variable and the `[0]` indexing in the line above aren't necessary and will be removed.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,480 @@
...
+    if rc == FFI_DECLINED then
+        return nil, nil, "key not found"
+    end
+
+    return ffi_str(cvalue[0].data, cvalue[0].len), tonumber(ccas[0])
+end
+
+
+local function set_kv_value(zone, key, value, cas)
+    if type(key) ~= "string" then
```
casimiro, line 206: Missing test for empty string. Will add later.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,480 @@
...
+local function new_histogram_buffer()
+    local hbs = C.ngx_wa_ffi_shm_max_histogram_size()
+    local hbuf = new_zero_filled_buffer("u_char [?]", hbs)
+
+    return hbuf, hbs
+end
+
+
+local function parse_cmetric(cmetric)
+    if type(cmetric) ~= "cdata" then
```
casimiro, line 268: This check isn't really needed since `parse_cmetric` is a local function.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,475 @@
...
+    end
+
+    if opts ~= nil then
+        if type(opts) ~= "table" then
+            error("opts must be a table", 2)
+        end
+
+        if opts.prefix ~= nil then
+            if type(opts.prefix) ~= "boolean" then
+                error("opts.prefix must be a boolean")
```
casimiro, line 330: Add `2` as argument.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,475 @@
...
+end
+
+
+local function get_metric(zone, metric_id)
+    if type(metric_id) ~= "number" then
+        error("metric_id must be a number", 2)
+    end
+
+    local mbuf, mbs = new_metric_buffer()
+    local hbuf, hbs = new_histogram_buffer()
```
casimiro, line 301: Reuse these buffers and simply erase them.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,475 @@
...
+
+
+local function new_metric_buffer()
+    local mbs = C.ngx_wa_ffi_shm_one_slot_metric_size()
+    local mbuf = new_zero_filled_buffer("u_char [?]", mbs)
+
+    return mbuf, mbs
+end
+
+local function new_histogram_buffer()
```
casimiro, line 258: Remove extra declaration.

docs/PROXY_WASM.md
```diff
@@ -749,6 +749,11 @@ payloads.
    eviction mechanism when the allocated memory slab is full:
     - `proxy_enqueue_shared_queue`
 
+5. In ngx_wasm_module, metric names are prefixed with `pw.filter_name.`, where
+   `filter_name` represents the name assigned to the filter in which the metric
+   is defined.
+   In Envoy, however, metric names are prefixed with `wasmcustom.`.
```
thibaultcha, line 755: I think it may be difficult for users to understand when specifically this difference matters. Am I right that it only matters to them when retrieving metrics by name through the FFI `get_metric_by_name`? Since in the SDK metrics are retrieved by id only?

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -96,7 +92,8 @@ histogram_bin(ngx_wa_metrics_t *metrics, ngx_wa_metrics_histogram_t *h,
             return b;
         }
 
-        if (b->upper_bound == 0) {
+        if (b->upper_bound == NGX_MAX_UINT32_VALUE) {
+            i++;
```
thibaultcha, line 96: This change feels like it is something else than exposing histograms to `ngx_wa_metrics_get`, can you elaborate?

docs/PROXY_WASM.md
casimiro: That's right.

I included this paragraph after inspecting Envoy's response on its admin metrics endpoint `/metrics` and seeing that it was exposing metrics with their internal prefixing, e.g. `wasmcustom.waf_filter.tx.total`, instead of just `waf_filter.tx.total`. 

But I hadn't noticed the `/metrics/prometheus` admin endpoint which properly exposes the metric names without their internal prefixing, e.g. `waf_filter_tx_total`.

Yeah, I think the user can be spared the details of internal prefixing if they never have to deal with them.

Since this information is only relevant to the FFI `get_metric_by_name` users , perhaps we should remove the paragraph entirely from `PROXY_WASM.md` and leave it as a comment in the FFI code.

src/common/metrics/ngx_wa_histogram.c
casimiro: Indeed, this change is not required to have histograms exposed in `ngx_wa_metrics_get`.

This change is meant to make the code clearer and more closely aligned with the design of histograms' bins -- it doesn't actually change the behaviour of the iteration itself.

Histograms are created with 5 bins, the first one initialized with  `NGX_MAX_UINT32_VALUE` as upper bound:
`[(0, NGX_MAX_UINT32_VALUE), (0, 0), (0, 0), (0, 0), (0, 0)]`.

When a new value, e.g. 1, is added, the first bin is shifted to the right to make room for the new bin:
`[(1, 2), (0, NGX_MAX_UINT32_VALUE), (0, 0), (0, 0), (0, 0)]`.

In this design, the upper bound `NGX_MAX_UINT32_VALUE` always signals the end of the initialized bins -- the bins to the right aren't initialized and shouldn't be visited. 

Previously, the code was iterating the bins until it found one whose `upper_bound` was 0. That worked, but it should really be iterating over the initialized bins only.

docs/PROXY_WASM.md
thibaultcha: Hmmm yes, I think we should document this in he FFI. As for removing it from our public PROXY_WASM.md documentation, I agree, but I also think we should keep track of all our internal differences with other host SDK implementations. Maybe a `.md` document in `assets/`?

src/common/metrics/ngx_wa_histogram.c
thibaultcha: Ok got it, thanks for the explanation!

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -257,3 +258,218 @@ ngx_http_wasm_ffi_set_host_properties_handlers(ngx_http_request_t *r,
...
 }
 #endif
+
+
+void
+ngx_wa_ffi_shm_lock(ngx_wa_shm_t *shm)
+{
+    ngx_wa_assert(shm);
+
+    ngx_wa_shm_lock(shm);
```
thibaultcha, line 268: Not a blocker for this PR, I think we can merge it as such, but reading this makes me realize that all locking APIs in OpenResty also come with a timeout handler that automatically releases the lock on expiration. Given our FFI API is not directly exposed to end-users this is probably fine for now granted we review & ensure correct usage, but a timeout handler and API option could become necessary in the future.

casimiro: Hm, that's interesting. Could you point me to one of these APIs?

I don't see such a timeout handler being employed to release locks in the `shdict` library. See this [excerpt](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_shdict.c#L524-L578) from the `ngx_http_lua_shdict_get_keys`, for example.

thibaultcha: Only when the lock is being explicitly manipulated by the end-user, not when it is part of the OpenResty codebase itself. E.g. the [semaphore API](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.md#wait), or other lua-resty libraries built on top of locking mechanisms (and exposing these locks to the end-user). Hence as we are not advertising our FFI lib to end-users (only to Gateway developers), I think we can skip this safeguard, but we should still keep it in mind wen reviewing Gateway code using this FFI.

casimiro: Oh, got it, thanks!

docs/PROXY_WASM.md
thibaultcha: I think I will do this locally to speed up the merge - if you haven't started yet.

casimiro: Yeah, I have. I messaged you privately to confirm it would be okay to:
- drop the doc commit entirely
- amend the FFI commit with the comment about prefixing
- add a separate commit containing a new `assets/INTERNAL_DIFFERENCES.md` file.

The changes are ready.

thibaultcha: Ok let's do it then!

PR #590, opened 2024-09-09
chore(util/release) update OpenSSL download link

casimiro:


casimiro:
The new OpenSSL version doesn't build on centos7 due to a dependency on a version of a Perl module not yet available on the platform.

A new [commit](https://github.com/openssl/openssl/commit/210dc9a50dfd99caa1cf7c3d2fa42850124b1bbc) removing that dependency has been pushed and should be available in the next release.

thibaultcha:
I think we can merge this as-is, granted we remove the temp commit - can you? In the meantime no release until a new OpenSSL version comes out with the fix then...

casimiro:
@thibaultcha dropped the temporary commit.

PR #591, opened 2024-09-19
[merge] feat/metrics_lua_interface

thibaultcha:
@casimiro Please have a look especially at the `wip()` commit and let me know if everything is still looking good. I ended up moving the document back to `docs/` but not linking anything to it as it's so minor.


lib/resty/wasmx/shm.lua
```diff
@@ -149,144 +155,163 @@ local _metric_type_set = {
...
+    ctx.i = ctx.i + 1
+
+    return key
 end
 
 
-local function iterate_keys(zone, page_size)
-    if page_size ~= nil and type(page_size) ~= "number" then
-        error("page_size must be a number", 2)
+local function shm_iterate_keys(zone, opts)
```
casimiro, line 215: I think grouping optional arguments into `opts` is an overkill here considering that `key_iterator` is likely to only be configured in two ways: page size and yield/not_yield.

I must admit, the additional complexity required for validating grouped optional arguments annoys me.

lib/resty/wasmx/shm.lua
```diff
@@ -149,144 +155,163 @@ local _metric_type_set = {
...
 
-    local rc = C.ngx_wa_ffi_shm_get_kv_value(shm, cname, cvalue, ccas)
+    local rc = C.ngx_wa_ffi_shm_kv_get(shm, cname, cvalue, ccas)
     if rc == FFI_DECLINED then
-        return nil, nil, "key not found"
+        return nil
+    end
+
+    if rc ~= FFI_OK then
+        return nil, "unknown error"
```
casimiro, line 307: I don't understand why you changed the logic around the return codes. 

`ngx_wa_ffi_shm_kv_get` returns whatever `ngx_wa_shm_kv_get_locked` returns. 
`ngx_wa_shm_kv_get_locked` returns either `NGX_DECLINED` when a key is not found or `NGX_OK` when it is found.

casimiro: There's really no space for an unknown error.

lib/resty/wasmx/shm.lua
```diff
@@ -295,78 +320,186 @@ local function set_kv_value(zone, key, value, cas)
...
-    local rc = C.ngx_wa_ffi_shm_set_kv_value(shm, cname, cvalue, cas, written)
-    if rc == FFI_ERR then
-        return nil, "failed setting kv value, no memory"
+    local rc = C.ngx_wa_ffi_shm_kv_set(shm, cname, cvalue, cas, written)
+    if rc == FFI_ERROR then
+        return nil, "no memory"
+    end
+
+    if rc ~= FFI_OK then
+        return nil, "unknown error"
```
casimiro, line 341: The previous comment also applies here.
`ngx_wa_ffi_shm_kv_set` returns whatever `ngx_wa_shm_kv_set_locked` returns.
`ngx_wa_shm_kv_set_locked` returns either `NGX_ERROR` if there's not enough memory for the new value or `NGX_OK` if the new value is successfully set.

lib/resty/wasmx/shm.lua
```diff
@@ -295,78 +320,186 @@ local function set_kv_value(zone, key, value, cas)
...
+
+    if rc == FFI_ABORT then
+        return nil, "name too long"
+    end
+
+    if rc ~= FFI_OK then
+        -- FFI_BUSY: unreachable
+        return nil, "unknown error"
+    end
+
```
casimiro, line 380: The check `rc ~= FFI_OK` above isn't really needed.

I noticed you changed `ngx_wa_metrics_define` to return `NGX_BUSY` if the function is called with an unknown metric type. I can't really see the benefit of discriminating the particular types of bad usage here. Specially considering that `ngx_wa_metrics_define` will never receive a bad metric type.

I think the existing error code convention is enough to capture errors that the user needs to be aware of, e.g. metric name too long.

lib/resty/wasmx/shm.lua
```diff
@@ -295,78 +320,186 @@ local function set_kv_value(zone, key, value, cas)
...
 
-    local rc = C.ngx_wa_ffi_shm_get_metric(metric_id, nil,
-                                           _mbuf, _mbs, _hbuf, _hbs)
+    local rc = C.ngx_wa_ffi_shm_metric_get(metric_id, nil,
+                                           _mbuf, _mbs,
+                                           _hbuf, _hbs)
     if rc == FFI_DECLINED then
-        return nil, "metric not found"
+        return nil
+    end
```
casimiro, line 483: We should return an error message `metric not found`.

lib/resty/wasmx/shm.lua
```diff
@@ -295,78 +320,186 @@ local function set_kv_value(zone, key, value, cas)
...
+    value = value and value or 1
+
+    local rc = C.ngx_wa_ffi_shm_metric_increment(metric_id, value)
+    if rc == FFI_DECLINED then
+        return nil, "metric not found"
+    end
+
+    if rc ~= FFI_OK then
+        return nil, "unknown error"
+    end
```
casimiro, line 403: Ditto error check not needed

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -304,126 +288,126 @@ ngx_wa_ffi_shm_setup_zones(ngx_wa_ffi_shm_setup_zones_handler_pt setup_handler)
...
-ngx_wa_ffi_shm_get_kv_value(ngx_wa_shm_t *shm,
-                            ngx_str_t *k, ngx_str_t **v, uint32_t *cas)
+ngx_wa_ffi_shm_kv_get(ngx_wa_shm_t *shm, ngx_str_t *k, ngx_str_t **v,
+    uint32_t *cas)
 {
-    unsigned   unlock = 0;
     ngx_int_t  rc;
 
-    ngx_wa_assert(shm && k && v && cas);
+    ngx_wa_assert(shm->type == NGX_WA_SHM_TYPE_KV);
```
casimiro, line 373: It's not clear to me the criteria we should use to decide if something should be asserted.

I thought we had agreed to only check for pointers validity.
The current assertion seems unnecessary as `ngx_wa_ffi_shm_kv_get` is bound to only be called with `NGX_WA_SHM_TYPE_KV` shms from `shm.lua`. That being said, `shm.lua` also guarantees that `ngx_wa_ffi_shm_get_kv_value` will only receive valid pointers.

I also don't think these assertions add any substantial amount of documentation to what these functions do.

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -304,126 +288,126 @@ ngx_wa_ffi_shm_setup_zones(ngx_wa_ffi_shm_setup_zones_handler_pt setup_handler)
...
 
-    if (!ngx_wa_shm_locked(shm)) {
-        ngx_wa_shm_lock(shm);
-        unlock = 1;
-    }
+    ngx_wa_shm_lock(shm);
```
casimiro, line 375: This change prevents the user from being able to retrieve values while iterating over the keys of an `shm` 
```
shm.kv:lock()

for key in shm.kv:iterate_keys() do
    local value = shm.kv:get(key)
end

shm.kv:unlock()

File: src/common/lua/ngx_wasm_lua_ffi.c [hunk 9]
```diff
@@ -304,126 +288,126 @@ ngx_wa_ffi_shm_setup_zones(ngx_wa_ffi_shm_setup_zones_handler_pt setup_handler)
...
     ngx_int_t  rc;
 
-    ngx_wa_assert(shm && k && v && written);
+    ngx_wa_assert(shm->type == NGX_WA_SHM_TYPE_KV);
 
-    if (!ngx_wa_shm_locked(shm)) {
-        ngx_wa_shm_lock(shm);
-        unlock = 1;
-    }
+    ngx_wa_shm_lock(shm);
```
casimiro, line 393: Ditto not allowing the user to set values while iterating over the keys of an shm.

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -304,126 +288,126 @@ ngx_wa_ffi_shm_setup_zones(ngx_wa_ffi_shm_setup_zones_handler_pt setup_handler)
...
 static void
-shm_kv_retrieve_keys(ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel,
-                     ngx_uint_t start,  ngx_uint_t limit,
-                     ngx_str_t **keys, ngx_uint_t *total)
+retrieve_keys(ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel,
+    ngx_uint_t limit, ngx_uint_t start_idx, ngx_uint_t *cur_idx,
+    ngx_str_t **keys)
 {
-    ngx_wa_shm_kv_node_t *n = (ngx_wa_shm_kv_node_t *) node;
+    ngx_wa_shm_kv_node_t  *n = (ngx_wa_shm_kv_node_t *) node;
```
casimiro, line 295: Good catch!

casimiro:
Thank you for improving the PR, @thibaultcha! 

I left comments about error handling, locking/unlocking in (get/set)_kv when already locked and a note on optional arguments for `iterate_keys()`.

lib/resty/wasmx/shm.lua
```diff
@@ -0,0 +1,577 @@
...
+    ffi_fill(_mbuf, _mbs)
+    ffi_fill(_hbuf, _hbs)
+
+    name = (opts and opts.prefix == false) and name or "lua." .. name
+
+    local cname = ffi_new("ngx_str_t", { data = name, len = #name })
+
+    local rc = C.ngx_wa_ffi_shm_metric_get(0, cname, _mbuf, _mbs, _hbuf, _hbs)
+    if rc == FFI_DECLINED then
+        return nil
```
casimiro, line 526:
```suggestion
        return nil, `metric not found`
```

lib/resty/wasmx/shm.lua
thibaultcha: Throughout our codebase we assert return codes of all underlying C APIs for defensive reasons, future-proofing, etc. The FFI should be no exception. I originally had `assert(rc == FFI_OK)` but switched to a less-harsh standard error. Assertions provide a nice stacktrace to investigate but standard errors may avoid the error-out for non-critical code paths. This really is all about being defensive so I thought a non-intrusive error would be better.

lib/resty/wasmx/shm.lua
thibaultcha: As written in the comment the code path is unreachable but the code is defensive and explicit.

src/common/lua/ngx_wasm_lua_ffi.c
thibaultcha: If a pointer is invalid, the function will crash later and it will be obvious in the produced coredump. Having assertions that contain several conditions within them also require the coredump to be opened in the first place (since the code line won't be enough to determine the failing condition). It can also be argued that according to the code in `shm.lua`, `k`, `v,` and `cas` would never be invalid when reaching this code path, so as per that argument the original assertion was unnecessary in the first place as well. As it stands today it isn't really necessary either but it makes the function more explicit to the reader so I kept it from another function that checked the type (assertions were inconsistent) and consistently applied it to all functions.

lib/resty/wasmx/shm.lua
thibaultcha: This is a single return value API just like shm `get()`.

src/common/lua/ngx_wasm_lua_ffi.c
thibaultcha: Is it really a good idea to do that? If so there should be tests for this, at the minimum.

src/common/lua/ngx_wasm_lua_ffi.c
thibaultcha: Fair - will also need tests.

lib/resty/wasmx/shm.lua
casimiro: Got it. It did occur to me that it could come from defensive programming but it wasn't clear if that was standard we should follow throughout the FFI layer.

Thank you for the explanation!

src/common/lua/ngx_wasm_lua_ffi.c
thibaultcha: No strong feelings on this from my end, we can go back to super-explicit assertions too.

lib/resty/wasmx/shm.lua
casimiro: Got it. Thanks for the explanation!

lib/resty/wasmx/shm.lua
thibaultcha: I am thinking of switching these back to assertions, and perhaps only having these assertions enabled in debug mode (which iirc is possible to check somehow as I think I did it previously).

src/common/lua/ngx_wasm_lua_ffi.c
casimiro: It hasn't occurred to me that having multiple conditions in a single assert makes it harder to distinguish which one was false. Good point. 

Thank you for the thoughtful explanation.

src/common/lua/ngx_wasm_lua_ffi.c
casimiro: I included a test case for this situation in the original [PR](https://github.com/Kong/ngx_wasm_module/pull/582/files#diff-c41e29444a09e4c8bb9f78a60119c76724751ac81101d489497b794b758dfbe2R15-R53).

Do you want me to push a commit adding a new version of this test?

thibaultcha: I recall this test but the note was for myself. I must also say, I remembered this usage in the test but as I stated in our call, it was actually not properly isolated (testing multiple new APIs together in the same suite/test). A proper test case would have been separate from the sanity test and explicitly state the purpose of the test (e.g. "can call get() while iterating).

lib/resty/wasmx/shm.lua
casimiro: Hm, `iterate_keys(1024, false)` does look more obscure. Fair point.

src/common/lua/ngx_wasm_lua_ffi.c
casimiro: I'm sorry, I misunderstood the note.

Thank you for clarifying the convention over test isolation and modularity. I wasn't mindful of those conventions while writing those tests, but will be next time.

PR #592, opened 2024-09-20
chore(ci) fix local httpbin instance in macos

casimiro:


casimiro:
The issue seems to be fixed already.

PR #595, opened 2024-09-26
feat(metrics/histogram) sample sum support

casimiro:


t/04-openresty/ffi/shm/024-metrics_get_by_name.t
```diff
@@ -78,7 +78,7 @@ prefix: pw.hostcalls.*
 --- response_body
 c1: {type="counter",value=13}
 g1: {type="gauge",value=1}
-h1: {type="histogram",value={{count=1,ub=128},{count=0,ub=4294967295}}}
+h1: {sum=100,type="histogram",value={{count=1,ub=128},{count=0,ub=4294967295}}}
```
thibaultcha, line 81: Can we add this new sum member to the proper (non-FFI) histogram tests as well? Probably `proxy_record_metric.t`. Additionally while adding it there, probably in the same existing test (`record_metrics() histogram`), can we add more than one value to the bin, so we can ensure the sum is indeed a sum. As it stands, this test with `sum=100` would pass even if the C code is modified to do `h->sum = n`, instead of `h->sum += n`.

casimiro: Certainly, but the test I updated [here](https://github.com/Kong/ngx_wasm_module/pull/595/files#diff-2551c8baa35c917cc0891013cc4a9a3643b7bf9f7d31a238d63140bd70a6168eR840) seems to fit these requirements. Did I miss something?.

I did oversee an update to a test in `proxy_record_metric_edge_cases.t`, though. Will push an update soon.

thibaultcha: Yes, that is good.

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -222,6 +226,7 @@ ngx_wa_metrics_histogram_get(ngx_wa_metrics_t *metrics, ngx_wa_metric_t *m,
 
     for (i = 0; i < slots; i++) {
         h = m->slots[i].histogram;
+        out->sum += h->sum;
```
thibaultcha, line 229: Could this be `out->sum = h->sum` to be more explicit to the reader?

casimiro: `out->sum` is meant to accumulate the sum of each slot's histogram, hence the `+=`.
If we change value attribution to `=`, `out->sum` will no longer hold the accumulated sum but the histogram's sum of the last slot.

thibaultcha: Oh yes my bad, I misread this for the FFI function...

thibaultcha (approved):

PR #608, opened 2024-10-14
feat(metrics) histograms with user-defined bins

casimiro:
This feature is only available through the FFI.

If a user provides a list of numbers when defining a histogram, those numbers will be used as the upper-bounds of the histogram's bins.

For instance, the following code:

```
shm.metrics:define("h", shm.metrics.HISTOGRAM, { bins = { 1, 3, 5 } })
```

Creates a histogram with bins `[0, 1]`, `(1, 3]`, `(3, 5]` and `(5, Inf+]`.


lib/resty/wasmx/shm.lua
```diff
@@ -355,12 +366,37 @@ local function metrics_define(zone, name, metric_type)
...
+            if type(opts.bins) ~= "table" then
+                error("opts.bins must be a table", 2)
+            end
+
+            if #opts.bins > HISTOGRAM_MAX_BINS then
+                local err = "opts.bins must have up to %d numbers"
+                error(str_fmt(err, HISTOGRAM_MAX_BINS), 2)
+            end
+
+            sort(opts.bins)
```
thibaultcha, line 387: There is no guarantee up to this point that the `opts.bins` elements are numbers, or that they are positive either. We should probably add some checks?

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -77,7 +77,25 @@ histogram_grow(ngx_wa_metrics_t *metrics, ngx_wa_metrics_histogram_t *h,
 
 
 static ngx_wa_metrics_bin_t *
-histogram_bin(ngx_wa_metrics_t *metrics, ngx_wa_metrics_histogram_t *h,
+histogram_custom_bin(ngx_wa_metrics_histogram_t *h, ngx_uint_t n)
+{
+    uint16_t               i = 0;
```
thibaultcha, line 82:
```suggestion
    size_t               i;
```

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -160,13 +178,23 @@ histogram_log(ngx_wa_metrics_t *metrics, ngx_wa_metric_t *m, uint32_t mid)
...
+    uint16_t                      j = 0, n_bins = NGX_WA_METRICS_BINS_INIT;
+    ngx_wa_histogram_type_e       h_type = NGX_WA_HISTOGRAM_LOG2;
     ngx_wa_metrics_histogram_t  **h;
 
+    if (bins) {
+        if (cn_bins > NGX_WA_METRICS_BINS_MAX) {
+            return NGX_ABORT;
+        }
+
+        n_bins = cn_bins + 1;
```
thibaultcha, line 194: Maybe a comment to describe what this is about?

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -205,8 +242,20 @@ ngx_wa_metrics_histogram_record(ngx_wa_metrics_t *metrics, ngx_wa_metric_t *m,
...
+        b = histogram_log2_bin(metrics, h, n, &m->slots[slot].histogram);
+        break;
+    case NGX_WA_HISTOGRAM_CUSTOM:
+        b = histogram_custom_bin(h, n);
+        break;
+    default:
+        return NGX_ERROR;
+    }
+
+    if (b) {
```
thibaultcha, line 256: So if `b` is `NULL` (which can only happen with custom bins?), what does it mean? A comment maybe?

src/common/metrics/ngx_wa_metrics.c
```diff
@@ -316,7 +318,7 @@ ngx_wa_metrics_define(ngx_wa_metrics_t *metrics, ngx_str_t *name,
     m->type = type;
 
     if (type == NGX_WA_METRIC_HISTOGRAM) {
-        rc = ngx_wa_metrics_histogram_add_locked(metrics, m);
+        rc = ngx_wa_metrics_histogram_add_locked(metrics, bins, n_bins,  m);
```
thibaultcha, line 321:
```suggestion
        rc = ngx_wa_metrics_histogram_add_locked(metrics, bins, n_bins, m);
```

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -160,13 +178,23 @@ histogram_log(ngx_wa_metrics_t *metrics, ngx_wa_metric_t *m, uint32_t mid)
...
 
 ngx_int_t
-ngx_wa_metrics_histogram_add_locked(ngx_wa_metrics_t *metrics,
-    ngx_wa_metric_t *m)
+ngx_wa_metrics_histogram_add_locked(ngx_wa_metrics_t *metrics, uint32_t *bins,
+    uint16_t cn_bins, ngx_wa_metric_t *m)
 {
     size_t                        i;
-    static uint16_t               n_bins = NGX_WA_METRICS_BINS_INIT;
+    uint16_t                      j = 0, n_bins = NGX_WA_METRICS_BINS_INIT;
```
thibaultcha, line 185: `j` should be declared as per `i` (type and line) since it's a loop control variable.

lib/resty/wasmx/shm.lua
```diff
@@ -355,12 +366,37 @@ local function metrics_define(zone, name, metric_type)
         error(err, 2)
     end
 
+    local cbins = nil
```
thibaultcha, line 369:
```suggestion
    local cbins
```

lib/resty/wasmx/shm.lua
```diff
@@ -355,12 +366,37 @@ local function metrics_define(zone, name, metric_type)
...
+    local n_bins = 0
+
+    if opts ~= nil then
+        if type(opts) ~= "table" then
+            error("opts must be a table", 2)
+        end
+
+        if opts.bins ~= nil then
```
thibaultcha, line 377: Should we skip this branch when the metric type is not histogram, but `opts.bins` is given? We can probably silently ignore it (as this already does, but it still creates the bins array buffer).

lib/resty/wasmx/shm.lua
casimiro: Good point. I'll add validation for the contents of `opts.bins` as well.

src/common/metrics/ngx_wa_histogram.c
casimiro: I opted for declaring `i` as `uint16_t` because we compare it with `h->n_bins` which is also `uint16_t`. I'm not sure it's safe to compare `size_t`, 8 bytes long, with `uint16_t`, 4 bytes long. In fact, in [one](https://github.com/Kong/ngx_wasm_module/actions/runs/11329386762/job/31504815070#step:9:7399) of the CI runs gcc complained about comparing `size_t` with `uint16_t` somewhere else I had misdeclared types.

src/common/metrics/ngx_wa_histogram.c
casimiro: Fair.

src/common/metrics/ngx_wa_histogram.c
casimiro: `b` can't really be `NULL`. I added the check anticipating that the compiler would complain about dereferencing a pointer that *theoretically* can be `NULL` (e.g., passing a histogram with `n_bins=0` to `histogram_custom_bin` would make it return a `NULL` pointer).

I'll test removing the check.

src/common/metrics/ngx_wa_metrics.c
casimiro: Nice catch!

src/common/metrics/ngx_wa_histogram.c
thibaultcha: We already do so in existing code?

src/common/metrics/ngx_wa_histogram.c
thibaultcha: Then an assertion is more appropriate; we also might want to add a similar assertion to the custom_bin function, which initializes `b = NULL` and then runs the loop. If we add an assertion before the final `return b` we show the developer that the function cannot return `NULL`.

lib/resty/wasmx/shm.lua
casimiro: Well observed. I'll make the branch to be skipped when the metric type isn't histogram.

src/common/metrics/ngx_wa_histogram.c
thibaultcha: imho, it is almost always better to make the loop variables a larger type (hence all our loop variables are `size_t`), since we can always cast smaller types into larger ones, and if the variable we are comparing against ever changes types one day (unlikely here but this is a general rule I follow), the loop variable will always be large enough.

casimiro: Cool, I like declaring `i` as `size_t` and casting `uint16_t` to `size_t` before comparison.

src/common/metrics/ngx_wa_histogram.c
casimiro: Good point. I'll use assertions to communicate the validity of the pointers.

t/04-openresty/ffi/shm/020-metrics_define.t
```diff
@@ -20,10 +20,13 @@ __DATA__
         access_by_lua_block {
             local shm = require "resty.wasmx.shm"
 
+            local bins = { 1, 3, 5 }
```
thibaultcha, line 23: Let's also add a test with unsorted bins - unless I missed it?

casimiro: Good point. As privately discussed, we are forbidding unsorted bins.

lib/resty/wasmx/shm.lua
thibaultcha: Can we remove the call to `sort` now?

casimiro: Oh, my bad.

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -177,7 +208,16 @@ ngx_wa_metrics_histogram_add_locked(ngx_wa_metrics_t *metrics,
         }
 
         (*h)->n_bins = n_bins;
-        (*h)->bins[0].upper_bound = NGX_MAX_UINT32_VALUE;
+        (*h)->h_type = h_type;
+
+        if (bins) {
+            /* user-defined set of bins */
+            for (/* void */; j < (size_t) n_bins - 1; j++) {
```
thibaultcha, line 215: Why don't we reset `j` here?

casimiro: That's a bug, glad you catch it!

src/common/metrics/ngx_wa_histogram.c
```diff
@@ -160,13 +180,24 @@ histogram_log(ngx_wa_metrics_t *metrics, ngx_wa_metric_t *m, uint32_t mid)
...
 {
-    size_t                        i;
-    static uint16_t               n_bins = NGX_WA_METRICS_BINS_INIT;
+    size_t                        i, j = 0;
+    uint16_t                      n_bins = NGX_WA_METRICS_BINS_INIT;
+    ngx_wa_histogram_type_e       h_type = NGX_WA_HISTOGRAM_LOG2;
     ngx_wa_metrics_histogram_t  **h;
 
+    if (bins) {
+        if (cn_bins > NGX_WA_METRICS_BINS_MAX) {
```
thibaultcha, line 192: One last thing before merge: a logarithmic histogram can contain up to 18 bins as per the existing design, but this specific condition and the extra bin on line 197 mean that a custom binning histogram can contain up to 19 bins (18 defined + the MAX_UINT32 bin). Is this correct?

thibaultcha: I made this change in the merge PR. The change makes the documentation easier to read and understand: "all histograms can only have 18 bins"

casimiro: Thanks for addressing it in the merge PR.

PR #617, opened 2024-10-24
chore(metrics) drop 'mid' argument from `ngx_wa_metrics_histogram_record`

casimiro:
The argument was only used when logging the histogram in debug mode. Since our tests don't assert on the actual value of 'mid', we don't really need to have it printed in logs.

Also, `ngx_wa_metrics_record` already prints the mid of the metric being updated.


thibaultcha:
Nice, always nice to keep things tidy!

PR #618, opened 2024-10-28
misc(metrics) space as metadata separator in metric name

casimiro:
WasmX metrics defined in Proxy-Wasm filters have their names prefixed by `pw.{a_filter}` where `{a_filter}` is the name of the filter where the metric was defined, e.g. `pw.a_filter.a_counter`.

When the Gateway serializes a WasmX metric defined in a Proxy-Wasm filter it needs to extract the filter name from the metric name. The filter name is used to load potential label patterns associated with the filter which are then used to extract potential labels from the metric name.

This PR proposes changing the separator in metric names from the dot to the space (e.g. `pw a_filter a_counter`) and forbidding space-like characters in WASM modules names; which simplifies the task of extracting the filter name from a metric name.

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -38,9 +38,9 @@ gauge, or a histogram.
 ## Name Prefixing
 
 To avoid naming conflicts between Proxy-Wasm filters, the name of a metric is
-always prefixed with: `pw.{filter_name}.{metric_name}`. This means that a metric
-named `a_counter` inserted by `a_filter` will have its name stored as:
-`pw.a_filter.a_counter`.
+always prefixed with space-separated metadata: `pw {filter_name} `. This means
+that a metric named `a_counter` inserted by `a_filter` will have its name stored
+as: `pw a_filter a_counter`.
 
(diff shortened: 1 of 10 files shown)


docs/METRICS.md
```diff
@@ -38,9 +38,9 @@ gauge, or a histogram.
 ## Name Prefixing
 
 To avoid naming conflicts between Proxy-Wasm filters, the name of a metric is
-always prefixed with: `pw.{filter_name}.{metric_name}`. This means that a metric
-named `a_counter` inserted by `a_filter` will have its name stored as:
-`pw.a_filter.a_counter`.
+always prefixed with space-separated metadata: `pw {filter_name} `. This means
```
thibaultcha, line 41: Reading this I realized space might not be the ideal character; it's easy to miss the trailing space in docs describing `pw {filter_name} `. Would it be fine if we used colons like `pw:{filter_name}:`? I opened #619 for merge if that's alright.

casimiro: Indeed, colons are more easily spotted than spaces.

PR #620, opened 2024-10-29
chore(ci) cache httpbin docker image

casimiro:
Also prevents dnsmasq from being started on installation.

### DIFF
--- a/.github/actions/load-docker-image/action.yml
+++ b/.github/actions/load-docker-image/action.yml
@@ -1,0 +1,52 @@
+name: load-docker-image
+decription: |
+  Pull a docker image and cache it.
+
+inputs:
+  image:
+    required: true
+    type: string
+  cache_dir:
+    required: false
+    type: string
+    default: "$GITHUB_WORKSPACE/.docker-images-cache"
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Create cache directory
+      shell: bash
+      run: mkdir -p ${{ inputs.cache_dir }}
+
+    - name: Setup cache
+      uses: actions/cache@v4
+      with:
+          path: ${{ inputs.cache_dir }}
+          key: pulled-docker-images
+
+    - name: Setup cache path
+      id: cache_setup
+      shell: bash
+      run: |
+        filename="$(echo ${{ inputs.image }} | sed 's/\//_/g').tar"
+        cache_path="${{ inputs.cache_dir }}/$filename"
+
+        if [ -f "$cache_path" ]; then
+          echo "cache_hit=true" >> $GITHUB_OUTPUT
+        else
+          echo "cache_hit=false" >> $GITHUB_OUTPUT
+        fi
+
+        echo "cache_path=$cache_path" >> $GITHUB_OUTPUT
+
+    - name: Pull Docker image
+      if: steps.cache_setup.outputs.cache_hit == 'false'
+      shell: bash
+      run: |
+        docker pull ${{ inputs.image }}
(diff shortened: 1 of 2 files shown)


thibaultcha:
Fantastic! Should we also cache and load the `nginx` image before the "Build httpbin-proxy image" step? This way we also avoid the `FROM nginx` step to pull it from the hub as well.

casimiro:
Good idea! I'll include `nginx` image in the cache.

casimiro:
On a second thought, `docker/build-push-action` is configured to cache its build environment which makes me conclude that it's already preventing `nginx` image from being pulled whenever the action `setup-httpbin-server` is ran. 

Do you think it's worth caching `nginx` image regardless of that?

casimiro:
Yes, we could have an additional `docker/build-push-action` step and Dockerfile to build an image that simply encapsulates httpbin's. Although less verbose in the `setup-httpbin-server` action, this seems a too implicit approach for caching an image that's simply pulled.

Maybe the ideal approach would be moving the new caching steps from `setup-httpbin-server` to a separate action `cache-docker-image` (or something like that) and have `setup-httpbin-server` simply invoking it. That way `setup-httpbin-server` doesn't get too cluttered and `cache-docker-image` represents an explicit intent that we can reuse in our workflows.

What do you think?

thibaultcha:
Ok, let's do the separate action and encapsulate the httpbin image like we do the nginx one. Although let's name this separate action `load-docker-image`, because that seems to be the core usage of this new action: make an image available for use, and cache it as an implementation detail. Sounds good?

.github/actions/load-docker-image/action.yml
```diff
@@ -0,0 +1,52 @@
...
+  Pull a docker image and cache it.
+
+inputs:
+  image:
+    required: true
+    type: string
+  cache_dir:
+    required: false
+    type: string
+    default: "~/.docker-images-cache"
```
thibaultcha, line 12: Let's only use GHA-defined variables instead of `~`; the list is: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables.

.github/actions/setup-httpbin-server/action.yml
```diff
@@ -70,6 +79,13 @@ runs:
         context: .
         cache-from: type=gha
         cache-to: type=gha,mode=max
+
+    - name: Load httpbin docker image
+      if: ${{ !env.ACT }}
+      uses: ./.github/actions/load-docker-image
+      with:
+        image: "kennethreitz/httpbin"
```
thibaultcha, line 87: Let's also do it for the nginx image like we said?

.github/actions/load-docker-image/action.yml
casimiro: Good call!

.github/actions/setup-httpbin-server/action.yml
casimiro: I thought we had decided not to explicitly cache the nginx image since it will be implictly cached by `docker/build-push-action` on the `Build httpbin-proxy image` step.

thibaultcha: Ok. Well, is there a way that both the nginx and httpbin images be cached and loaded in the same way? Even if we have to create our own Dockerfile for the httpbin one. The goal I was trying to aim for was consistency in the loading/caching of all our Docker images used in CI.

thibaultcha: Ok no need, they're too different. Merged!

PR #621, opened 2024-10-29
feat(ffi) fails explicitly if the user tries to set a value on a locked shm

casimiro:
This change also significantly reduces the execution time of the `t/04-openresty/ffi/shm` test suite.


lib/resty/wasmx/shm.lua
```diff
@@ -346,6 +346,10 @@ local function shm_kv_set(zone, key, value, cas)
         return nil, "no memory"
     end
 
+    if rc == FFI_ABORT then
+        return nil, "cannot set value if the shm is already locked"
```
thibaultcha, line 350: In our FFI API we need very short error strings for two reasons: users can compare them to catch the error:
```lua
local ok, err = shm:set(...)
if err == "no memory" then

end
```

Or users will also concatenate it to an error string of their own choosing:
```lua
if not ok then
    return nil, "cannot set value in shm: " .. err
end
```

My suggestion for this specific error is to just say "locked".

thibaultcha:
Ultimately I fail to understand how this helps; the current promise is that `set()` will be executed as soon as the lock is available. But if we error-out while the shm is locked, then users have to embed their `set()` call inside of some loop construct so as to retry on locked shm, this feels very subpar. What is the benefit of erroring-out on locked shms?

thibaultcha:
Yeah ok I see. To me this is a misuse - I'll elaborate. We have lock/unlock surrounding the shm_set call already:
```c
// ngx_wasm_lua_ffi.c
ngx_wa_shm_lock(shm);
rc = ngx_wa_shm_kv_set_locked(shm, k, v, cas, written);
ngx_wa_shm_unlock(shm);
```
I also considered this when working on the merge of the keys iterator PR and it bugged me that users could get themselves into this situation (locking the shm manually + calling a function that also locks the shm, resulting in the deadlock). Ultimately I thought it was fine to treat it as a misuse of a low-level API. It'd be great to catch it (distinguishing where the lock came from), but I'm not sure the trouble is worth it when once we have FFI lib documentation, we will mention how the locks work for `shm.lua` calls, and there add a notice to the `shm:lock()` API that it is a low-level utility that's necessary for `iterate_keys()` but can lead to edge-cases if used for other purposes. Again we could catch it in the code if we put in some effort, but since the only users are the Gateway contributors...

thibaultcha:
Now I understand you wanted to catch that case of misuse, specifically. I should have checked `ngx_wa_shm_locked` sooner as I see we can totally rely on it to verify it is the current worker who holds the lock! So this should indeed catch the case more easily than I first assumed.

thibaultcha:
At first I thought the intent was to prevent `set()` from going through when iterating (as per the comment) to which I said we do require the lock before writing anyway so we did not need it. But what this does really is catching a call to `set()` when `lock()` was called (not when another `set()` from another worker is holding the lock).

casimiro:
> what this check does really is catching a call to set() when lock() was called in the same worker

Oh, only now I noticed that I haven't made it explicit that I was addressing the case where the mutex is locked by the same process attempting to lock it. I'll update the added comment to make it explicit about the case being addressed.

lib/resty/wasmx/shm.lua
casimiro: I like the simplicity of just returning `locked`.

I'm a bit concerned though that it might not be enough to communicate to the developer that `shm:set()` shouldn't be called while the shm is being iterated over using `shm:iterate_keys()` (I understand that users can explicitly lock the shm without iterating over it, but since there's no point in doing so I'm ignoring this case).

I chose a more verbose `err` because I didn't expected the user to check on its value, as it's just a message communicating an API misuse.

It's not like the message I proposed is doing a substantially better job communicating the misuse though, I agree. And I also find it difficult to communicate the actual cause of the error in a relatively short message.

I'll change it to `locked`.

src/common/lua/ngx_wasm_lua_ffi.c
```diff
@@ -368,8 +368,8 @@ ngx_int_t
 ngx_wa_ffi_shm_kv_get(ngx_wa_shm_t *shm, ngx_str_t *k, ngx_str_t **v,
     uint32_t *cas)
 {
-    ngx_int_t   rc;
-    unsigned    unlock = 0;
+    ngx_int_t  rc;
+    unsigned   unlock = 0;
```
thibaultcha, line 372: Note the spacing is correct here: 2 empty spaces + an extra space for potential pointers.

thibaultcha:
Manually merged :+1:

PR #625, opened 2024-11-15
feat(proxy-wasm) invoke 'on_http_call_response' on dispatch failures

### LINKED ISSUE #622: How to handle dispatch_http_call/other timeout
Even though you can specify timeout in dispatch_http_call Rust SDK, it doesn't seem to return any Status and the function stops instead. Same case for invalid upstream.

thibaultcha:
Invoke `on_http_call_response` on dispatch connection failures. This allows catching failures such as:

- timeout
- broken connection
- resolver failures
- TLS handshake failures
- Any other failures *after the connection has attempted to be established*

Dispatch failures occurring *before* a connection has attempted to be established will not trigger `on_http_call_response` as they are already supposed to trigger a Wasm exception (i.e. trap).

When a dispatch connection failure occurs, `on_http_call_response` is invoked with: `on_http_call_response(call_id, 0, 0, 0)` (i.e. no header, no body, no trailers). Calls to retrieve the response `:status` will return `None`. A user may triage the connection failure by querying the `:dispatch_status` pseudo-header:

```rust
fn on_http_call_response(
    &mut self,
    token_id: u32,
    nheaders: usize,
    body_size: usize,
    ntrailers: usize,
) {
    let dispatch_status = self.get_http_call_response_header(":dispatch_status");

    match dispatch_status.as_deref() {
        Some("timeout") => {},
        Some("broken connection") => {},
        Some("tls handshake failure") => {},
        Some("resolver failure") => {},
        Some("reader failure") => {},
        Some(s) => info!("dispatch_status: {}", s),
        None => {}
    }

    self.resume_http_request()
}
```

The socket status `"bad argument"` also exists but since the connection has not been attempted yet, it won't show up during `on_http_call_response` and is for internal use only (i.e. could be added to the socket error log for example).

Fix #622.

---

TODO

- ~[ ] contribute `proxy_get_status()` for HTTP dispatch calls to proxy-wasm-rust-sdk~
- [x] a directive to not log socket errors
- [x] document connection error handling & Envoy disparities in our docs

### DIFF
--- a/t/03-proxy_wasm/hfuncs/131-proxy_dispatch_http_timeouts.t
+++ b/t/03-proxy_wasm/hfuncs/131-proxy_dispatch_http_timeouts.t
@@ -14,7 +14,7 @@ use t::TestWasmX;
 our $ExtResolver = $t::TestWasmX::extresolver;
 our $ExtTimeout = $t::TestWasmX::exttimeout;
 
-plan_tests(4);
+plan_tests(6);
 run_tests();
 
 __DATA__
@@ -32,14 +32,19 @@ qq{
         wasm_socket_connect_timeout 1ms;
 
         proxy_wasm hostcalls 'test=/t/dispatch_http_call \
-                              host=google.com';
+                              host=google.com \
+                              on_http_call_response=echo_response_headers';
         echo ok;
     }
 }
---- response_body
-ok
+--- response_body_like
+:dispatch_status: timeout
 --- error_log eval
-qr/(\[error\]|Uncaught RuntimeError|\s+).*?dispatch failed: tcp socket - timed out connecting to \".*?\"/
+[
+    qr/(\[error\]|Uncaught RuntimeError|\s+).*?dispatch failed: tcp socket - timed out connecting to \".*?\"/,
+    qr/\*\d+ .*? on_http_call_response \(id: \d+, status: , headers: 0, body_bytes: 0/,
+    qr/dispatch_status: timeout/
+]
 --- no_error_log
 [crit]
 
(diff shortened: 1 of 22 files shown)


hishamhm:
> ```rust
>     let dispatch_status = self.get_http_call_response_header(":dispatch_status");
> ```

I assume that this is in an interim solution until we can get something like `get_http_call_status` (which would call the `proxy_get_status` hostcall in `HttpContext`) added to the proxy-wasm-rust-sdk, right? I can work with that.

thibaultcha:
> I assume that this is in an interim solution until we can get something like `get_http_call_status` (which would call the `proxy_get_status` hostcall in `HttpContext`) added to the proxy-wasm-rust-sdk, right? I can work with that.

Yes, and I looked into this and it was unclear how to contribute this at first sight, so I logged it for later. It looks to me like `get_http_call_status` could cause a spec breaking change or at least it could be a host breaking change for existing implementations. At the moment, `proxy_get_status` (the hostcall) is only used in [ctx.get_grpc_status](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/hostcalls.rs#L981) and it worries me that it only receives return pointers, as the host must do some trickery to know where it is being invoked from ("are we executing http callback or grpc callback?"). The problem is that both gRPC and HTTP callbacks can belong to the HTTP filter context trait (`HttpContext`), so a host must restrict usage to within the callback and be able to distinguish which type of callback is currently running. The ideal would even be to pass it the HTTP/gRPC `token_id` argument so it could be queried from anywhere and not just the callback. That said for now it looks to me like Envoy (or proxy-wasm-cpp-host) is NYI on this hostcall.

t/03-proxy_wasm/hfuncs/131-proxy_dispatch_http_timeouts.t
```diff
@@ -31,6 +31,45 @@ qq{
...
         wasm_socket_connect_timeout 1ms;
 
+        proxy_wasm hostcalls 'test=/t/dispatch_http_call \
+                              host=google.com \
+                              on_http_call_response=echo_response_headers';
+        echo ok;
+    }
+}
+--- response_body_like
+:dispatch_status: timeout
```
thibaultcha, line 41: @hishamhm This version will support `get_http_headers` and return the `:dispatch_status` pseudo-header in the table

hishamhm:
@thibaultcha I've just tested this with the Gateway and caught call errors in the filter successfully!

PR #626, opened 2024-11-18
feat(proxy-wasm) foreign function support

casimiro:
### Foreign Function Support

The Proxy-Wasm [spec](https://github.com/proxy-wasm/spec/tree/main/abi-versions/v0.2.1#foreign-function-interface-ffi) defines the function `proxy_call_foreign_function` and the callback `proxy_on_foreign_function` that filter developers can use to invoke host specific functions, a.k.a. foreign functions, and receive callbacks.

A foreign function invoked with `proxy_call_foreign_function` may return its value immediately -- as part of the returned value of the `proxy_call_foreign_function` call; or it can return it later, writing it to the `FOREIGN_FUNCTION_ARGUMENTS` [buffer](https://github.com/proxy-wasm/spec/tree/main/abi-versions/v0.2.1#foreign-function-interface-ffi), and invoking `proxy_on_foreign_function` with an id identifying the function initially called.

This PR adds support for the mechanism described above; and although the spec doesn't restrict when foreign functions can be called, in `ngx_wasm_module` they cannot be invoked from `proxy_on_configure` or `proxy_on_vm_start`.

### DNS resolution

This PR also exposes the Lua DNS resolver through the `resolve_lua` foreign function.

This function expects the name to be resolved as its single argument. If the name can be resolved without performing any IO, the resolved address is put in the returned value of `proxy_call_foreign_function`.

If the resolver needs to forward the resolution request to a DNS server, the resolved address and the name itself are written to the `FOREIGN_FUNCTION_ARGUMENTS` buffer and the `proxy_on_foreign_function` callback is invoked with `function_id` 0, as soon as the address is returned from the server.

### TODO

- [x] `on_tick` support
- [x] update docs
- [x] general review

### DIFF
--- a/config
+++ b/config
@@ -144,7 +144,8 @@ NGX_WASMX_DEPS="\
     $ngx_addon_dir/src/common/metrics/ngx_wa_metrics.h \
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm.h \
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_maps.h \
-    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_properties.h"
+    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_properties.h \
+    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.h"
 
 NGX_WASMX_SRCS="\
     $ngx_addon_dir/src/ngx_wasmx.c \
@@ -160,8 +161,9 @@ NGX_WASMX_SRCS="\
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm.c \
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_host.c \
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_maps.c \
+    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_util.c \
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_properties.c \
-    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_util.c"
+    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.c"
 
 # wasm
 
--- a/src/common/proxy_wasm/ngx_proxy_wasm.h
+++ b/src/common/proxy_wasm/ngx_proxy_wasm.h
@@ -78,6 +78,7 @@ typedef enum {
     NGX_PROXY_WASM_STEP_DONE,
     NGX_PROXY_WASM_STEP_TICK,
     NGX_PROXY_WASM_STEP_DISPATCH_RESPONSE,
+    NGX_PROXY_WASM_STEP_FOREIGN_CALLBACK,
 } ngx_proxy_wasm_step_e;
 
 
@@ -124,7 +125,7 @@ typedef enum {
     NGX_PROXY_WASM_BUFFER_GRPC_RECEIVE_BUFFER = 5,
     NGX_PROXY_WASM_BUFFER_VM_CONFIGURATION = 6,
     NGX_PROXY_WASM_BUFFER_PLUGIN_CONFIGURATION = 7,
-    NGX_PROXY_WASM_BUFFER_CALL_DATA = 8,
+    NGX_PROXY_WASM_BUFFER_FOREIGN_FUNCTION_ARGUMENTS = 8,
 } ngx_proxy_wasm_buffer_type_e;
 
 
@@ -147,6 +148,11 @@ typedef enum {
 } ngx_proxy_wasm_metric_type_e;
 
 
+typedef enum {
+    NGX_PROXY_WASM_FOREIGN_RESOLVE_LUA = 0,
+} ngx_proxy_wasm_foreign_function_e;
+
+
 typedef struct ngx_proxy_wasm_ctx_s  ngx_proxy_wasm_ctx_t;
 typedef struct ngx_proxy_wasm_filter_s  ngx_proxy_wasm_filter_t;
 typedef struct ngx_proxy_wasm_exec_s  ngx_proxy_wasm_exec_t;
@@ -154,43 +160,64 @@ typedef struct ngx_proxy_wasm_instance_s  ngx_proxy_wasm_instance_t;
 #ifdef NGX_WASM_HTTP
 typedef struct ngx_http_proxy_wasm_dispatch_s  ngx_http_proxy_wasm_dispatch_t;
 #endif
+typedef struct ngx_proxy_wasm_foreign_call_s  ngx_proxy_wasm_foreign_call_t;
 typedef ngx_str_t  ngx_proxy_wasm_marshalled_map_t;
 
 
 typedef struct {
-    ngx_queue_t                        busy;
-    ngx_queue_t                        free;
-    ngx_queue_t                        sweep;
-    ngx_pool_t                        *pool;
+    ngx_queue_t                         busy;
+    ngx_queue_t                         free;
+    ngx_queue_t                         sweep;
+    ngx_pool_t                         *pool;
 } ngx_proxy_wasm_store_t;
 
 
 typedef struct {
-    ngx_str_t                          log_prefix;
-    ngx_log_t                         *orig_log;
-    ngx_proxy_wasm_exec_t             *pwexec;
+    ngx_str_t                           log_prefix;
+    ngx_log_t                          *orig_log;
+    ngx_proxy_wasm_exec_t              *pwexec;
 } ngx_proxy_wasm_log_ctx_t;
 
 
-struct ngx_proxy_wasm_exec_s {
-    ngx_uint_t                         root_id;
-    ngx_uint_t                         id;
-    ngx_uint_t                         index;
-    ngx_uint_t                         tick_period;
-    ngx_rbtree_node_t                  node;
-    ngx_proxy_wasm_err_e               ecode;
-    ngx_pool_t                        *pool;
-    ngx_log_t                         *log;
-    ngx_proxy_wasm_log_ctx_t           log_ctx;
-    ngx_proxy_wasm_ctx_t              *parent;
-    ngx_proxy_wasm_filter_t           *filter;
-    ngx_proxy_wasm_instance_t         *ictx;
-    ngx_proxy_wasm_store_t            *store;
-    ngx_event_t                       *ev;
+typedef enum {
+    NGX_PROXY_WASM_DISPATCH_HTTP_CALL,
+    NGX_PROXY_WASM_DISPATCH_FOREIGN_CALL,
+} ngx_proxy_wasm_dispatch_op_e;
+
+
+typedef struct {
+    ngx_queue_t                          q;     /* stored by caller */
+    ngx_proxy_wasm_dispatch_op_e         type;
+
+    union {
 #ifdef NGX_WASM_HTTP
-    ngx_http_proxy_wasm_dispatch_t    *dispatch_call;  /* swap pointer for host functions */
+        ngx_http_proxy_wasm_dispatch_t  *http;
 #endif
-    ngx_queue_t                        dispatch_calls;
+        ngx_proxy_wasm_foreign_call_t   *foreign;
+    } call;
+} ngx_proxy_wasm_dispatch_op_t;
+
+
+struct ngx_proxy_wasm_exec_s {
+    ngx_uint_t                          root_id;
+    ngx_uint_t                          id;
+    ngx_uint_t                          index;
+    ngx_uint_t                          tick_period;
+    ngx_rbtree_node_t                   node;
+    ngx_proxy_wasm_err_e                ecode;
+    ngx_pool_t                         *pool;
+    ngx_log_t                          *log;
+    ngx_proxy_wasm_log_ctx_t            log_ctx;
+    ngx_proxy_wasm_ctx_t               *parent;
+    ngx_proxy_wasm_filter_t            *filter;
+    ngx_proxy_wasm_instance_t          *ictx;
+    ngx_proxy_wasm_store_t             *store;
+    ngx_event_t                        *ev;
+    ngx_queue_t                         dispatch_ops;
+    ngx_proxy_wasm_foreign_call_t      *foreign_call;   /* swap pointer for host functions */
+#ifdef NGX_WASM_HTTP
+    ngx_http_proxy_wasm_dispatch_t     *dispatch_call;  /* swap pointer for host functions */
+#endif
 
     /* flags */
 
@@ -414,8 +441,10 @@ ngx_int_t ngx_proxy_wasm_resume(ngx_proxy_wasm_ctx_t *pwctx,
     ngx_wasm_phase_t *phase, ngx_proxy_wasm_step_e step);
 ngx_proxy_wasm_err_e ngx_proxy_wasm_run_step(ngx_proxy_wasm_exec_t *pwexec,
     ngx_proxy_wasm_step_e step);
-ngx_uint_t ngx_proxy_wasm_dispatch_calls_total(ngx_proxy_wasm_exec_t *pwexec);
-void ngx_proxy_wasm_dispatch_calls_cancel(ngx_proxy_wasm_exec_t *pwexec);
+ngx_uint_t ngx_proxy_wasm_dispatch_ops_total(ngx_proxy_wasm_exec_t *pwexec);
+void ngx_proxy_wasm_dispatch_ops_cancel(ngx_proxy_wasm_exec_t *pwexec);
+ngx_uint_t ngx_proxy_wasm_foreign_calls_total(ngx_proxy_wasm_exec_t *pwexec);
+void ngx_proxy_wasm_foreign_calls_cancel(ngx_proxy_wasm_exec_t *pwexec);
 
 
 /* host handlers */
--- a/src/common/proxy_wasm/ngx_proxy_wasm.h
+++ b/src/common/proxy_wasm/ngx_proxy_wasm.h
@@ -188,9 +188,9 @@ struct ngx_proxy_wasm_exec_s {
     ngx_proxy_wasm_store_t            *store;
     ngx_event_t                       *ev;
 #ifdef NGX_WASM_HTTP
-    ngx_http_proxy_wasm_dispatch_t    *call;  /* swap pointer for host functions */
+    ngx_http_proxy_wasm_dispatch_t    *dispatch_call;  /* swap pointer for host functions */
 #endif
-    ngx_queue_t                        calls;
+    ngx_queue_t                        dispatch_calls;
 
     /* flags */
 
@@ -232,19 +232,19 @@ struct ngx_proxy_wasm_ctx_s {
     size_t                                        req_body_len;
     ngx_str_t                                     authority;
     ngx_str_t                                     scheme;
-    ngx_str_t                                     path;              /* r->uri + r->args */
-    ngx_str_t                                     start_time;        /* r->start_sec + r->start_msec */
-    ngx_str_t                                     upstream_address;  /* 1st part of ngx.upstream_addr */
-    ngx_str_t                                     upstream_port;     /* 2nd part of ngx.upstsream_addr */
-    ngx_str_t                                     connection_id;     /* r->connection->number */
-    ngx_str_t                                     mtls;              /* ngx.https && ngx.ssl_client_verify */
-    ngx_str_t                                     root_id;           /* pwexec->root_id */
-    ngx_str_t                                     call_status;       /* dispatch response status */
-    ngx_str_t                                     response_status;   /* response status */
+    ngx_str_t                                     path;                  /* r->uri + r->args */
+    ngx_str_t                                     start_time;            /* r->start_sec + r->start_msec */
+    ngx_str_t                                     upstream_address;      /* 1st part of ngx.upstream_addr */
+    ngx_str_t                                     upstream_port;         /* 2nd part of ngx.upstsream_addr */
+    ngx_str_t                                     connection_id;         /* r->connection->number */
+    ngx_str_t                                     mtls;                  /* ngx.https && ngx.ssl_client_verify */
+    ngx_str_t                                     root_id;               /* pwexec->root_id */
+    ngx_str_t                                     dispatch_call_status;  /* dispatch response status */
+    ngx_str_t                                     response_status;       /* response status */
 #if (NGX_DEBUG)
-    ngx_str_t                                     worker_id;         /* ngx_worker */
+    ngx_str_t                                     worker_id;             /* ngx_worker */
 #endif
-    ngx_uint_t                                    call_code;
+    ngx_uint_t                                    dispatch_call_code;
     ngx_uint_t                                    response_code;
 
     /* host properties */
@@ -259,9 +259,9 @@ struct ngx_proxy_wasm_ctx_s {
 
     /* flags */
 
-    unsigned                                      main:1;            /* r->main */
-    unsigned                                      init:1;            /* can be utilized (has no filters) */
-    unsigned                                      ready:1;           /* filters chain ready */
+    unsigned                                      main:1;                /* r->main */
+    unsigned                                      init:1;                /* can be utilized (has no filters) */
+    unsigned                                      ready:1;               /* filters chain ready */
     unsigned                                      req_headers_in_access:1;
 };
 
--- a/src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.c
@@ -1,0 +1,305 @@
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+#if (NGX_WASM_LUA)
+#include <ngx_wasm_lua_resolver.h>
+#endif
+#include <ngx_proxy_wasm_foreign_call.h>
+
+
+void
+ngx_proxy_wasm_foreign_call_destroy(ngx_proxy_wasm_foreign_call_t *call)
+{
+    ngx_pfree(call->pwexec->pool, call);
+}
+
+
+#if (NGX_WASM_HTTP)
+#if (NGX_WASM_LUA)
+static void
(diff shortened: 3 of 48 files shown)


src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -147,50 +148,58 @@ typedef enum {
...
+    ngx_log_t                          *log;
+    ngx_proxy_wasm_log_ctx_t            log_ctx;
+    ngx_proxy_wasm_ctx_t               *parent;
+    ngx_proxy_wasm_filter_t            *filter;
+    ngx_proxy_wasm_instance_t          *ictx;
+    ngx_proxy_wasm_store_t             *store;
+    ngx_event_t                        *ev;
 #ifdef NGX_WASM_HTTP
-    ngx_http_proxy_wasm_dispatch_t    *call;  /* swap pointer for host functions */
+    ngx_http_proxy_wasm_dispatch_t     *call;  /* swap pointer for host functions */
```
thibaultcha, line 198: dispatch_call

src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -147,50 +148,58 @@ typedef enum {
...
 #endif
-    ngx_queue_t                        calls;
+    ngx_queue_t                         calls;
```
thibaultcha, line 200: dispatch_calls

src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -147,50 +148,58 @@ typedef enum {
...
+    ngx_proxy_wasm_foreign_cb_t        *fcallback;
```
thibaultcha, line 201: foreign_call

src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -147,50 +148,58 @@ typedef enum {
...
+    ngx_queue_t                         fcallbacks;
```
thibaultcha, line 202: foreign_calls

src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -415,6 +424,9 @@ ngx_proxy_wasm_err_e ngx_proxy_wasm_run_step(ngx_proxy_wasm_exec_t *pwexec,
     ngx_proxy_wasm_step_e step);
 ngx_uint_t ngx_proxy_wasm_dispatch_calls_total(ngx_proxy_wasm_exec_t *pwexec);
 void ngx_proxy_wasm_dispatch_calls_cancel(ngx_proxy_wasm_exec_t *pwexec);
+ngx_uint_t ngx_proxy_wasm_foreign_callbacks_total(
```
thibaultcha, line 427: foreign_calls_total
...

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.c
```diff
@@ -0,0 +1,131 @@
...
+    size_t size)
+{
+    ngx_buf_t              *b;
+    ngx_chain_t            *cl;
+    ngx_proxy_wasm_exec_t  *pwexec = cb->pwexec;
+
+    ngx_wa_assert(pwexec);
+    ngx_wa_assert(size);
+
+    cl = ngx_alloc_chain_link(pwexec->pool);
```
thibaultcha, line 92: ngx_wasm_chain_get_free_buf

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.c
```diff
@@ -0,0 +1,131 @@
...
+    if (cl == NULL) {
+        return NGX_ERROR;
+    }
+
+    /* TODO: if size exceeds a threshold, split allocation into N buffers */
```
thibaultcha, line 97: ngx_wa_assert(len < NGX_MAX_ERRLEN)

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.c
```diff
@@ -0,0 +1,131 @@
...
+    cl->buf = b;
+    cl->next = NULL;
+
+    cb->args_out = cl;
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
```
thibaultcha, line 113: void

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+    ngx_http_request_t           *r;
+    ngx_resolver_ctx_t           *rslv_ctx;
+    ngx_wasm_socket_tcp_t        *sock;
+    ngx_proxy_wasm_exec_t        *pwexec;
+    ngx_proxy_wasm_foreign_cb_t  *cb;
+    ngx_wavm_ptr_t                p = 0;
+
+    pwexec = ngx_proxy_wasm_instance2pwexec(instance);
+
+    cb = ngx_proxy_wasm_foreign_callback_alloc(pwexec);
```
thibaultcha, line 1931: (inlined)

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.h
```diff
@@ -0,0 +1,29 @@
...
+#include <ngx_wasm.h>
+#include <ngx_proxy_wasm.h>
+#include <ngx_wasm_subsystem.h>
+
+
+struct ngx_proxy_wasm_foreign_cb_s {
+    ngx_queue_t                         q;
+    ngx_proxy_wasm_exec_t              *pwexec;
+#if (NGX_WASM_HTTP)
+    ngx_http_wasm_req_ctx_t            *rctx;
```
thibaultcha, line 14:
```suggestion
```

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+#if (NGX_HAVE_INET6)
+    case AF_INET6:
+          sin6 = (struct sockaddr_in6 *) rslv_ctx->addr.sockaddr;
+
+          *(p++) = sizeof(struct in6_addr);
+          p = ngx_cpymem(p, &sin6->sin6_addr, sizeof(struct in6_addr));
+
+          break;
+#endif
+
```
thibaultcha, line 1878:
```suggestion
```

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+    }
+
+    switch (sa_family) {
+#if (NGX_HAVE_INET6)
+    case AF_INET6:
+          sin6 = (struct sockaddr_in6 *) rslv_ctx->addr.sockaddr;
+
+          *(p++) = sizeof(struct in6_addr);
+          p = ngx_cpymem(p, &sin6->sin6_addr, sizeof(struct in6_addr));
+
```
thibaultcha, line 1875:
```suggestion
```

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+    cb = (ngx_proxy_wasm_foreign_cb_t *) sock->data;
+    p = buf;
+
+    ngx_memzero(buf, sizeof(buf));
+
+    if (rslv_ctx->state || !rslv_ctx->naddrs) {
+        p++;
+        goto not_found;
+    }
+
```
thibaultcha, line 1867: ngx_wa_assert(rslv_ctx->naddrs == 1) (perhaps?)

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+{
+    size_t                        s;
+    ngx_int_t                     rc;
+    ngx_buf_t                    *b;
+    ngx_http_request_t           *r;
+    ngx_resolver_ctx_t           *rslv_ctx;
+    ngx_wasm_socket_tcp_t        *sock;
+    ngx_proxy_wasm_exec_t        *pwexec;
+    ngx_proxy_wasm_foreign_cb_t  *cb;
+    ngx_wavm_ptr_t                p = 0;
```
thibaultcha, line 1927:
```suggestion
    ngx_wavm_ptr_t                p;
```

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+            != NGX_OK)
+        {
+            goto error;
+        }
+
+    } else {
+        cb->rctx = rctx;
+    }
+
+    sock = ngx_palloc(pwexec->pool, sizeof(ngx_wasm_socket_tcp_t));
```
thibaultcha, line 1951:
```suggestion
    sock = ngx_pcalloc(pwexec->pool, sizeof(ngx_wasm_socket_tcp_t));
```

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.c
```diff
@@ -0,0 +1,131 @@
...
+    ngx_pfree(cb->pwexec->pool, cb);
+}
+
+
+ngx_proxy_wasm_foreign_cb_t *
+ngx_proxy_wasm_foreign_callback_alloc(ngx_proxy_wasm_exec_t *pwexec)
+{
+    ngx_proxy_wasm_foreign_cb_t  *cb;
+
+    cb = ngx_palloc(pwexec->pool, sizeof(ngx_proxy_wasm_foreign_cb_t));
```
thibaultcha, line 22:
```suggestion
    cb = ngx_pcalloc(pwexec->pool, sizeof(ngx_proxy_wasm_foreign_cb_t));
```

+ inlined

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -1807,7 +1834,202 @@ ngx_proxy_wasm_hfuncs_increment_metric(ngx_wavm_instance_t *instance,
...
+    ngx_wavm_ptr_t                p = 0;
+
+    pwexec = ngx_proxy_wasm_instance2pwexec(instance);
+
+    cb = ngx_proxy_wasm_foreign_callback_alloc(pwexec);
+    if (cb == NULL) {
+        goto error;
+    }
+
+    cb->fcode = NGX_PROXY_WASM_FOREIGN_RESOLVE;
```
thibaultcha, line 1936: Let's try with a function pointer when all those routiens move to "dispatch_call.c"

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.c
```diff
@@ -0,0 +1,131 @@
...
+
+    b_size = b->end - b->start;
+
+    ngx_wa_assert(data->len <= b_size);
+
+    if (data->len <= b_size) {
+        b->last = ngx_cpymem(b->last, data->data, data->len);
+    }
+
+    /* TODO: data->len > b_size */
```
thibaultcha, line 128: move up to the assert

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.h
casimiro: Keeping `rctx` as it's not present in `ngx_proxy_wasm_exec_t` or `ngx_proxy_wasm_ctx_t`.

src/common/proxy_wasm/ngx_proxy_wasm_host.c
casimiro: The problem with using a function pointer here is that later we have to provide a function id to the Proxy-Wasm filter when we invoke `proxy_on_foreign_function` indicating the function originally executed by `call_foreign_function`.

As it is, `ngx_proxy_wasm_foreign_call_t` is simply a structure carrying the return value of a function invoked with `call_foreign_function` which yielded before returning.

thibaultcha: Hmmm ok then, sounds good (I now recall realizing that before our call the other day). Let's keep foreign call ops as an enum then.

src/common/proxy_wasm/ngx_proxy_wasm_foreign_callback.h
casimiro: Ah, thanks for clarifying!

thibaultcha: Ah right! So that settles it, sounds good then!

src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.c
```diff
@@ -0,0 +1,293 @@
...
+    ngx_wa_assert(size);
+
+    /* TODO: if size > NGX_MAX_ERROR_STR, split allocation into N buffers */
+
+    ngx_wa_assert(size <= NGX_MAX_ERROR_STR);
+
+    cl = ngx_wasm_chain_get_free_buf(pwexec->pool,
+                                     &rctx->free_bufs,
+                                     size, buf_tag,
+                                     rctx->sock_buffer_reuse);
```
thibaultcha, line 86: Great to see this all the updates; also to see such a shorter function. In this case though I think we have to hard-code the reuse argument to `1`, as `rctx->sock_buffer_reuse` is meant to be tied to TCP/UDP sockets. It is currently only used for HTTP dispatch sockets while all other calls to `ngx_wasm_chain_get_free_buf` are hard-coded as `1`. We could change this later or tie the hard-coded values to another directive (I doubt it's worth it, but we could), for now let's keep it that way and hard-code the reuse argument here?

thibaultcha: I can also make the change myself, but just an fyi

casimiro: Thanks for taking the time to explain that. Will had-code it to `1`.

src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.c
```diff
@@ -0,0 +1,293 @@
...
+          p = ngx_cpymem(p, &sin->sin_addr, sizeof(struct in_addr));
+    }
+
+not_found:
+
+    p = ngx_cpymem(p, rslv_ctx->name.data, rslv_ctx->name.len);
+    args.data = buf;
+    args.len = p - buf;
+
+    if (ngx_proxy_wasm_foreign_call_buffer_create(call, args.len) != NGX_OK) {
```
thibaultcha, line 174: When I look at the code of this `buffer_create` function now it seems to me like it should just be inlined. The core logic is just the buffer reuse, many of the lines in buffer_create can be avoided (e.g. the assertions), and this function is not reused anywhere else, so it feels like unnecessary redirection for the reader. I would say the same applies to `buffer_write` below as well.

casimiro: Yeah, well observed. I'll inline them.

casimiro:
Codecov is playing games with me. It's reporting lines that are definitely covered by tests as uncovered.

util/morestyle.pl
```diff
@@ -209,6 +209,10 @@
                 } elsif ($line =~ /^\s+[\+\-\*\/] [\s\w\(\)\+\-\*\/\.\-\>]/) {
                     # ignoring line with a portion of a wrapped expression
 
+                } elsif ($line =~ /^\s+(\w+(, )?)+\);/) {
+                    # ignoring line with arguments wrapped from previous line
+                    # TODO: ensure arguments' leading spaces are correct
```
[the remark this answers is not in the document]
casimiro: Splitting assignment from declaration has also been my default strategy so far.

The [case](https://github.com/Kong/ngx_wasm_module/pull/626/files#diff-ff18e71c99b245e5671673ec7af3e23d65224ef1b083ece8c4ccea163af5cabaR161-R171) that motivated the change involves an `#if` directive that would need to be duplicated if we were to separate assignment from declaration:

```c
{
    ngx_proxy_wasm_exec_t          *pwexec = ngx_proxy_wasm_instance2pwexec(
                                                 instance);
#if (NGX_WASM_LUA)
    /* remaining code */
```
vs
```c
{
#if (NGX_WASM_LUA)
    /* lua-only declarations */
#endif
    ngx_proxy_wasm_exec_t          *pwexec;

    pwexec = ngx_proxy_wasm_instance2pwexec(instance);

#if (NGX_WASM_LUA)
    /* remaining code */
```

Although I prefer the former, I don't strongly oppose the latter.

config
```diff
@@ -143,6 +143,7 @@ NGX_WASMX_DEPS="\
     $ngx_addon_dir/src/common/shm/ngx_wa_shm_queue.h \
     $ngx_addon_dir/src/common/metrics/ngx_wa_metrics.h \
     $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm.h \
+    $ngx_addon_dir/src/common/proxy_wasm/ngx_proxy_wasm_foreign_call.h \
```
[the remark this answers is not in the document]
casimiro: I did indeed think we were sorting by path, prefix and length, in order.

`ngx_proxy_wasm_util.c` is placed after `ngx_proxy_wasm_properties.c` which doesn't follow that sorting rule, though. That made me confused.

src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -147,50 +148,58 @@ typedef enum {
...
+    ngx_proxy_wasm_instance_t          *ictx;
+    ngx_proxy_wasm_store_t             *store;
+    ngx_event_t                        *ev;
 #ifdef NGX_WASM_HTTP
-    ngx_http_proxy_wasm_dispatch_t    *call;  /* swap pointer for host functions */
+    ngx_http_proxy_wasm_dispatch_t    *dispatch_call;  /* swap pointer for host functions */
 #endif
-    ngx_queue_t                        calls;
+    ngx_queue_t                        dispatch_calls;
+    ngx_proxy_wasm_foreign_call_t     *foreign_call;
```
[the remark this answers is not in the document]
casimiro: Thank you for reminding me about the comment!

src/http/ngx_http_wasm_util.c
```diff
@@ -725,7 +725,7 @@ ngx_http_wasm_cleanup_nop(void *data)
 #endif
 
 
-ngx_http_request_t *
+static ngx_http_request_t *
```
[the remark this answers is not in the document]
casimiro: It feels weird keeping it public if there's no code in main branch using it.
That being said, I see no issue in leaving it public, though.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
```diff
@@ -933,6 +915,15 @@ ngx_http_proxy_wasm_dispatch_resume_handler(ngx_wasm_socket_tcp_t *sock)
...
                                            NGX_PROXY_WASM_ACTION_PAUSE);
 
+    } else if (ngx_proxy_wasm_foreign_calls_total(pwexec)) {
+        ngx_log_debug0(NGX_LOG_DEBUG_WASM, pwexec->log, 0,
+                       "proxy_wasm foreign function callbacks pending...");
+
+        rc = NGX_AGAIN;
+        ngx_wasm_yield(&rctx->env);
+        ngx_proxy_wasm_ctx_set_next_action(pwexec->parent,
+                                           NGX_PROXY_WASM_ACTION_PAUSE);
```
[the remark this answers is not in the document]
casimiro: When I further modified this file to support yieldable foreign calls from `on_http_call_response` I became more apparent to me the duplication concern you've shared a few times.

Yup, combining the two queues into a single one will make the code simpler and more concise, thanks for suggesting!

t/lib/proxy-wasm-tests/hostcalls/Cargo.toml
```diff
@@ -8,7 +8,8 @@ edition = "2018"
 crate-type = ["cdylib"]
 
 [dependencies]
-proxy-wasm = "0.2"
+#proxy-wasm = "0.2"
+proxy-wasm = { git = "https://github.com/casimiro/proxy-wasm-rust-sdk.git", branch = "foreign-function-callback" }
```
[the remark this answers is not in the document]
casimiro: Yeah, I was hoping the changes in the upstream would be merged by the end of this PR.

On one hand, this workaround seems very subpar and makes me want to consider it a blocker.

On the other, that separate branch contains a single commit that's already been approved by the maintainers and should be part of the next release; and using this separate branch would allow us to ship foreign function support in time for the 3.9 release of the gateway.

t/03-proxy_wasm/hfuncs/contexts/140-proxy_foreign_function_resolve_lua.t
```diff
@@ -0,0 +1,169 @@
...
+--- config
+    location /t {
+        proxy_wasm context_checks 'on_log=call_resolve_lua';
+        return 200;
+    }
+--- error_log
+[error]
+can only call resolve_lua during "on_request_headers", "on_request_body", "on_dispatch_response", "on_tick", "on_foreign_function"
+--- no_error_log
+[crit]
```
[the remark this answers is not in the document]
casimiro: Yeah, as I was implementing support for calling `lua_resolver` from `on_http_call_response` I noticed that the code also allowed for calling `dispatch_http_call` from `on_foreign_call` and that `on_foreign_call` would be an additional check in the context tests for the hostcalls.

 There's a test asserting that filter developers can call the `lua_resolver` from the `on_http_call_response` callback; but tests not strictly related to the `lua_resolver` function  (e.g. calling `http_call` from `on_foreign_call` and the new `on_foreign_call` check in the context checks) were left to be addressed in a separate PR as this one is already rather large.
 
 I can include them in this PR. though; if you prefer that way.

I agree that there's value in keeping them separate.

t/03-proxy_wasm/hfuncs/contexts/140-proxy_foreign_function_resolve_lua.t
```diff
@@ -0,0 +1,169 @@
...
+plan_tests(4);
+run_tests();
+
+__DATA__
+
+=== TEST 1: proxy_wasm contexts - call_resolve_lua on_vm_start
+--- main_config
+    wasm {
+        module context_checks $TEST_NGINX_CRATES_DIR/context_checks.wasm 'call_resolve_lua';
+    }
```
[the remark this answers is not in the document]
casimiro: `--- wasm_modules` doesn't support passing the config string to the module directive.

t/03-proxy_wasm/hfuncs/141-proxy_resolve_lua.t
```diff
@@ -0,0 +1,188 @@
+# vim:set ft= ts=4 sts=4 sw=4 et fdm=marker:
```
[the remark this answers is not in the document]
casimiro: I had indeed considered this. Will move the file accordingly.

t/03-proxy_wasm/hfuncs/contexts/140-proxy_foreign_function_resolve_lua.t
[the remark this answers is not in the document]
thibaultcha: Ah yes, I thought there had to be a reason and missed this.

src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c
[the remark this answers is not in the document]
thibaultcha: It may be tricky to store both foreign and http dispatches in the same queue. Perhaps a single type with a `union` member could combine both HTTP calls and foreign calls, and that would be what is stored in the single dispatch_queue?

t/03-proxy_wasm/hfuncs/contexts/140-proxy_foreign_function_resolve_lua.t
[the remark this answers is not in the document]
thibaultcha: >  tests not strictly related to the lua_resolver function (e.g. calling http_call from on_foreign_call and the new on_foreign_call check in the context checks) were left to be addressed in a separate PR as this one is already rather large.

Let's add them in this PR, they belong to this feature and that way it will be done. The PR is not very large to me, it's alright.

src/common/proxy_wasm/ngx_proxy_wasm.h
```diff
@@ -124,7 +125,7 @@ typedef enum {
     NGX_PROXY_WASM_BUFFER_GRPC_RECEIVE_BUFFER = 5,
     NGX_PROXY_WASM_BUFFER_VM_CONFIGURATION = 6,
     NGX_PROXY_WASM_BUFFER_PLUGIN_CONFIGURATION = 7,
-    NGX_PROXY_WASM_BUFFER_CALL_DATA = 8,
+    NGX_PROXY_WASM_BUFFER_FOREIGN_FUNCTION_ARGUMENTS = 8,
```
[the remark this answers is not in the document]
casimiro: As we discussed privately, Rust and C++ SDKs name it `CALL_DATA` but the [spec](https://github.com/proxy-wasm/spec/tree/main/abi-versions/v0.2.1#buffers) refers to this buffer as `FOREIGN_FUNCTION_ARGUMENTS`. 

Although I appreciate the benefit of preserving consistence between the SDKs, in our case, `FOREIGN_FUCTION_ARGUMENTS` seems to better represent how the buffer is intended to be used.

src/common/proxy_wasm/ngx_proxy_wasm_host.c
```diff
@@ -146,6 +149,19 @@ ngx_proxy_wasm_get_buffer_helper(ngx_wavm_instance_t *instance,
...
 
+    case NGX_PROXY_WASM_BUFFER_FOREIGN_FUNCTION_ARGUMENTS:
+        {
+            ngx_proxy_wasm_foreign_call_t  *call = pwexec->foreign_call;
+
+            if (call == NULL) {
+                return NULL;
+            }
+
+            ngx_wa_assert(call->args_out);
```
thibaultcha, line 160: note to self

PR #627, opened 2024-11-21
feat(proxy-wasm) set querystring when setting ":path"

hishamhm:
* [tests(proxy-wasm) check that querystring works in dispatch_http_call](https://github.com/Kong/ngx_wasm_module/commit/67d48a344f01c37dddc12de398202a9893be5e73) - looks like setting the query passing `:path` to dispatch already worked; I added a test to verify this.
* [feat(proxy-wasm) set querystring when setting ":path"](https://github.com/Kong/ngx_wasm_module/commit/0bacf513bc9c6ea0766d6813bbdcaa3decd88392) - this one addresses the NYI, and includes some tests.
* [refactor(proxy-wasm) rename call->uri to call->path](https://github.com/Kong/ngx_wasm_module/commit/fd0fdcca407ae8a3f0dbaafe17483070b9559c75) - This third one is opinionated, feel free to keep or drop it. I just changed it to clarify that it matches `":path"` (that is, path including querystring), making the field name match the pseudo-header, as is the case with the other entries in `ngx_http_proxy_wasm_dispatch_s`.

### DIFF
--- a/t/03-proxy_wasm/hfuncs/111-proxy_set_http_request_header.t
+++ b/t/03-proxy_wasm/hfuncs/111-proxy_set_http_request_header.t
@@ -295,26 +295,135 @@ path: /test
 
 
 
-=== TEST 12: proxy_wasm - set_http_request_headers() cannot set ':path' with querystring (NYI)
+=== TEST 12: proxy_wasm - set_http_request_headers() can set ':path' with querystring
 --- wasm_modules: hostcalls
+--- http_config eval
+qq{
+    upstream test_upstream {
+        server unix:$ENV{TEST_NGINX_UNIX_SOCKET};
+    }
+
+    server {
+        listen unix:$ENV{TEST_NGINX_UNIX_SOCKET};
+
+        location / {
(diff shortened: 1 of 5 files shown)


t/03-proxy_wasm/hfuncs/111-proxy_set_http_request_header.t
```diff
@@ -295,26 +295,206 @@ path: /test
...
+(/test?foo=bár%20bla) (/test) (?) (foo=bár%20bla)
+--- error_log
+path: /test
+--- no_error_log
+[error]
+[crit]
+
+
+
+=== TEST 17: proxy_wasm - showcase that that path with invalid querystring passes through
```
hishamhm, line 462: I tried to make tests 16 and 17 into a single test using `--- request eval` and `["GET /set_by_proxy_wasm", "GET /raw_nginx"]`. That worked in isolation, but it messed up the `plan_tests` count.

thibaultcha:
Merged!

PR #629, opened 2024-11-26
tests(*) increase robustness of httpbin-reliant tests

thibaultcha:
We sometimes get a real 503 from httpbin.org as well. The tests still succeed in case of upstream errors.

### DIFF
--- a/t/03-proxy_wasm/hfuncs/130-proxy_dispatch_http.t
+++ b/t/03-proxy_wasm/hfuncs/130-proxy_dispatch_http.t
@@ -408,6 +408,7 @@ Needs IPv4 resolution + external I/O to succeed.
 Succeeds on:
 - HTTP 200 (httpbin.org/headers success)
 - HTTP 502 (httpbin.org Bad Gateway)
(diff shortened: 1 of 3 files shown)


casimiro:
Weird. The 503 would have come from the local httpbin instance that runs  during CI, not the real one hosted at httpbin.org, which is unexpected.

thibaultcha:
Sure, but the test is still considered successful.

thibaultcha:
In fact the error probably comes from our Nginx proxy in front. But again, we already expect this as per the test cases, this PR is not introducing anything new.

thibaultcha:
The setup of the [failing run](https://github.com/Kong/ngx_wasm_module/actions/runs/12022610649/job/33515116102) has a strange "setup local httpbin server" step in which we see:
```
Run docker rm -f httpbin httpbin_proxy
Error response from daemon: No such container: httpbin
Error response from daemon: No such container: httpbin_proxy
```
We can investigate it a bit. That said, iirc these special 5xx handling in these tests were there even before we had a local httpbin setup as httpbin.org was really giving us some 500s from time to time.

PR #655, opened 2024-12-14
fix(wavm) use PRIu64 for printf portability across architectures

### LINKED ISSUE #653: Wrong format for debugging on 64 bit platform
https://github.com/Kong/ngx_wasm_module/commit/73aaff4ba06e2648fa946c722c0849dd3a88ea7a#commitcomment-150327118

dev-null-undefined:
Fixes #653

### DIFF
--- a/src/wasm/vm/ngx_wavm.c
+++ b/src/wasm/vm/ngx_wavm.c
@@ -1151,7 +1151,7 @@ ngx_wavm_val_vec_set(wasm_val_vec_t *out, const wasm_valtype_vec_t *valtypes,
 
         case WASM_I64:
             ui64 = va_arg(args, uint64_t);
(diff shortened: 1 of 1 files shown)


thibaultcha:
Can we just use %lu and avoid anything too fancy? We strive for consistency and simplicity.

dev-null-undefined:
That would work only on 64bit platform and break on 32bit, which for debug printing is probably ok-ish, but I do not think that this is too complex for the benefit of being portable, ofc the call is yours.

thibaultcha:
It does seem like we require C99 already anyway so it should be good. To merge it we will need you to sign the CLA though.

thibaultcha:
Thank you!

PR #656, opened 2024-12-14
fix(proxy-wasm) pass hardcoded 0 as vm_configuration_size

### LINKED ISSUE #654: Wrong params for proxy_on_vm_start
https://github.com/proxy-wasm/spec/blob/main/abi-versions/v0.2.1/README.md#proxy_on_vm_start

The proxy_wasm_spec specifies that the second argument for proxy_on_vm_start sould be the VM config size

Since from my knowladge of the code you do not support VM config yet, it would make sense to hard code 0 there with maybe a comment /* TODO VM configuration */

https://github.com/Kong/ngx_wasm_module/blob/883b627dd4bdcdf5d0ce622855026087eee22592/src/common/proxy_wasm/ngx_proxy_wasm.c#L1216

dev-null-undefined:
pass hardcoded 0 instead of misleading constant

fixes #654

### DIFF
--- a/src/common/proxy_wasm/ngx_proxy_wasm.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm.c
@@ -1213,7 +1213,8 @@ ngx_proxy_wasm_create_context(ngx_proxy_wasm_filter_t *filter,
             rc = ngx_wavm_instance_call_funcref(ictx->instance,
                                                 filter->proxy_on_vm_start,
                                                 &rets,
-                                                rexec->id, rexec->root_id);
+                                                rexec->id,
+                                                filter->module->config.len);
             if (rc != NGX_OK || !rets->data[0].of.i32) {
                 ecode = NGX_PROXY_WASM_ERR_VM_START_FAILED;
                 goto error;


thibaultcha:
We do support VM config as discussed in the issue so let's pass the correct value instead. Such a PR will also need test cases.

dev-null-undefined:
> We do support VM config as discussed in the issue so let's pass the correct value instead. Such a PR will also need test cases.

My bad I missread your original comment on the issue and though you do not support VM configuration :D

thibaultcha:
Thanks for the update. To merge your PRs we will need you to sign our CLA as well as contribute the necessary regression tests when needed (this one needs it). For this test we will need to update the `001-on_root_phases.t` suite: duplicating the `TEST 1` with `on_vm_start without configuration` and `on_vm_start with configuration`. The second test needs to define a VM config in the `module` directive and assert the error log. It's very simple copy-pasting and many tests do this in the codebase already.

thibaultcha:
When dealing with tests, have a look at the developer documentation at https://github.com/Kong/ngx_wasm_module/blob/main/docs/DEVELOPER.md and look for the Makefile targets, for example `make test` or `make reindex`. The developer documentation is very detailed and should cover everything.

thibaultcha:
Done in #657, thanks!

PR #659, opened 2025-01-13
fix(proxy-wasm) catch get/set all scoped properties outside of requests

thibaultcha:
Scoped properties (e.g. `request.*`, `response.*`, `upstream.*`, ...) can only be read/written from within the context of a request. Only non-scoped properties (e.g. `worker_id`) can be accessed from root contexts.

Prior to this commit, only `ngx.*` and `wasmx.*` (host) properties were caught. Other scoped properties that did not map to either of those would cause a segfault.

We now catch all scoped properties (i.e. containing a `.` character) when getting/setting properties outside of a request.


dev-null-undefined:
I believe this patch is incorrect as for example the variable `plugin_name` will still cause segmentation fault when used in the on_tick "phase"

thibaultcha:
Nice catch thanks - the patch is fine, it's a different issue for `plugin_name` that request access. I'll send a PR.

thibaultcha:
See #667

PR #663, opened 2025-01-15
feat(proxy-wasm): Add new request property `request.is_subrequest`

### LINKED ISSUE #658: New property proposal - is_subrequest
New property named `is_subrequest` to enable wasm modules to recognize when request get's sliced etc.

Would you be willing to accept merge request adding this property? Or should I leave it just in my fork?

Thanks.

dev-null-undefined:
Introduce a new request property `request.is_subrequest` to indicate whether the current request is a subrequest of an existing request (e.g., in slicing scenarios).

Note: Although the property uses the `request.*` prefix, this is not an Envoy attribute.

Fixes #658

### DIFF
--- a/src/common/proxy_wasm/ngx_proxy_wasm_properties.c
+++ b/src/common/proxy_wasm/ngx_proxy_wasm_properties.c
@@ -172,6 +172,23 @@ get_request_time(ngx_proxy_wasm_ctx_t *pwctx, ngx_str_t *path,
 
 
 static ngx_int_t
+is_subrequest(ngx_proxy_wasm_ctx_t *pwctx, ngx_str_t *path,
+    ngx_str_t *value)
+{
+    ngx_http_wasm_req_ctx_t  *rctx = (ngx_http_wasm_req_ctx_t *) pwctx->data;
+    ngx_http_request_t       *r = rctx->r;
+
+    if (r == r->main) {
+        ngx_str_set(value, "false");
+    } else {
+        ngx_str_set(value, "true");
+    }
+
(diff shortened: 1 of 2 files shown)


thibaultcha:
Thank you. It looks like another test in the same suite is missing the new property and causing all these CI failures.

src/common/proxy_wasm/ngx_proxy_wasm_properties.c
```diff
@@ -171,6 +171,23 @@ get_request_time(ngx_proxy_wasm_ctx_t *pwctx, ngx_str_t *path,
...
+static ngx_int_t
+is_subrequest(ngx_proxy_wasm_ctx_t *pwctx, ngx_str_t *path,
+    ngx_str_t *value)
+{
+    ngx_http_wasm_req_ctx_t  *rctx = (ngx_http_wasm_req_ctx_t *) pwctx->data;
+    ngx_http_request_t       *r = rctx->r;
+
+    if (r == r->main) {
+        ngx_str_set(value, "false");
+    } else {
```
thibaultcha, line 183: The linter will also complain about this `else` which is missing the blank line above. We should actually do this logic like `connection.mtls`, and move the `static` `true/false` strings outside of the mtls getter so that we can have the same `value = r == r->main ? true : false` logic (see `connection.mtls` getter).

thibaultcha:
A rebase on main will also be necessary as it will fix the other CI issues.

thibaultcha:
Merged with edits in #666. Thank you!

PR #664, opened 2025-01-16
chore(ci): pin 3rd-party actions to specific commit hashes

Water-Melon:
KAG-6149


windmgc (approved):

ms2008 (approved):

thibaultcha:
Does this break our Dependabot automatic updates? If so, very much not inclined to merge this.

thibaultcha:
I was informed that Dependabot does support updating actions pinned with sha1 commits, in which case we are all good! Thank you.

